PR #12, opened 2021-04-22
Update naming of colors and spacing tokens

ryan-ludwig:
* Use "brand" instead of "green"
* Move to 0-1000 color names
* Change space naming from gap to numeric (tailwind style)

VUE-538, VUE-539


kv-tokens/primitives.json
```diff
@@ -5,33 +5,35 @@
...
-        "green": {
-            "110": "#26985D",
-            "105": "#28A062",
+        "gray": {
+            "800": "#212121",
+            "500": "#757575",
+            "300": "#C4C4C4",
+            "100": "#F5F5F5",
+            "50": "#fafafa"
+        },
```
mcstover, line 14: Just wanted to clarify, will these eventually become `gray-800` or `brand-550` in actual usage?

ryan-ludwig: @mcstover correct

mcstover (approved):

itseunsoo (approved):
Love the new naming conventions ❤️. With the grays, would it be bad practice to insert numbers in between if in the case we build out more shades? i.e. adding 200, 400, 600 values

ryan-ludwig:
> Love the new naming conventions ❤️. With the grays, would it be bad practice to insert numbers in between if in the case we build out more shades? i.e. adding 200, 400, 600 values

@itseunsoo Great! I think a nice benefit of this system is the ability to add more shades, even in small increments like 500, 550, 575, etc. so not a bad practice at all.

itseunsoo:
> > Love the new naming conventions ❤️. With the grays, would it be bad practice to insert numbers in between if in the case we build out more shades? i.e. adding 200, 400, 600 values
> 
> @itseunsoo Great! I think a nice benefit of this system is the ability to add more shades, even in small increments like 500, 550, 575, etc. so not a bad practice at all.

@ryan-ludwig That's awesome! Don't worry we won't add new grays every 50/25 increments and make a mess 😂, but very insightful for future reference if we need! Tyty

PR #21, opened 2021-04-29
Remove storybook script on kv-ui-elements package.json - running this…

eddieferrer:
… causes a weird infinite loop and never actually runs storybook


eddieferrer:
I wasn't able to run storybook at this level. I don't think this script should be here.

ryan-ludwig:
This should work if you have Node v15.13.0 or higher

ryan-ludwig:
I imagine this will catch other people up too though. I don't feel strongly about having the script at this level.

ryan-ludwig (approved):

PR #28, opened 2021-05-04
VUE-576 Tailwind prose plugin

ryan-ludwig:
Tailwind does a heavy reset on most elements, which means no default margins, list-styles etc. This is helpful when developing "app-like" pages.  For example, not having to reset `<ul>` every time you have a list of loan cards. Or not having the default bottom margin of a paragraph push down the padding of your loan-card component.

There are other times where this is very annoying, as you'd have to add `<p class="mb-4">` to every single paragraph in a blog post. Or maybe your content is coming in from Contentful and you only have raw HTML elements.

For the second use-case, Tailwind has a [prose typography plugin](https://github.com/tailwindlabs/tailwindcss-typography). It sets pretty defaults for standard HTML elements related to type. You use it by wrapping whatever content you want handled in a 
`<div class="prose"> ... your Contentful rich text here, or blog post, etc. </div>`.

This PR adds that plugin and overrides it's defaults to match the typography specced in the Figma Styleguide.

Note, our responsive type is handled in a different way, and we don't need to add their responsive size modifiers.
🙅‍♂️`<article class="prose prose-lg prose-xl etc..">`🙅‍♂️. 
👍`<article class="prose">`👍

-------

* Add `.text-small` class
* Refactor, breaking typography into new file
* Add prose plugin.
* Override some prose plugin defaults to match our typography.
* Add prose demos. Add .text-small demo


mcstover (approved):
Excited to see the comparison! But also, the convenience of the `prose` plugin is a really nice option!

ryan-ludwig:
Prose demo here: https://608ae61fe8204f002107b4e7-jzejjyybvd.chromatic.com/?path=/story/basestyling--prose-demo
No-prose demo: https://608ae61fe8204f002107b4e7-jzejjyybvd.chromatic.com/?path=/story/basestyling--no-prose-demo

You can see the markup in the "story" tab, next to accessibility.

emuvente (approved):
👍

BoulderBrains (approved):
👏

ryan-ludwig:
After sleeping on this, the amount of `false`s on the typography and overrides makes me wonder if we shouldn't just setup our own .prose styling using theirs as a starting point. Not worth doing today, but maybe in the future.

eddieferrer:
This is good bu I do wonder how it'll handle wrapping some of the richText from contentful... There's a couple of places where we are hacking the contentful richText to take <em> tags and highlight them green.

eddieferrer (approved):

PR #31, opened 2021-05-06
VUE-534 KvButton updates

ryan-ludwig:
* Adds the ripple effect on click or keyboard click
* Uses a `state` prop for adding disabled or loading states.
* Adds `type` prop for adding this button as a form submit

Stories: ~~https://608ae61fe8204f002107b4e7-nspambzcaq.chromatic.com/?path=/story/kvbutton--variant-primary~~ https://608ae61fe8204f002107b4e7-cavhquyhlt.chromatic.com/?path=/story/kvbutton--variant-primary

TODO:
* Design feedback
* `text-link` variant? Or should we make that a separate component?
* Replace clock with loading spinner once KvLoadingSpinner exists
* Add icons once KvIcon exists


kv-components/vue/KvButton.vue
```diff
@@ -30,44 +46,60 @@
...
+			type: String,
+			default: 'button',
+			validator(value) {
+				return ['button', 'submit', 'reset'].includes(value);
+			},
 		},
 		/**
 		 * Appearance of the button
-		 * `primary (default), secondary, link, danger`
+		 * `primary (default), secondary, danger, link, ghost`
```
ryan-ludwig, line 76: I don't love the name 'link' for this button appearance.  Any other ideas?
image

eddieferrer:
This is great, love that we can clean up some other VUE issues that have been around for a long time (adding a loading state, @click.native) 👍 

One thing I noticed, is that if a button has an href attribute, the styling changes. This looks like a subtle font change on most variants, and a pretty different styling on  the Secondary variant

ryan-ludwig:
> One thing I noticed, is that if a button has an href attribute, the styling changes. This looks like a subtle font change on most variants, and a pretty different styling on the Secondary variant

Good catch, thanks @eddieferrer !

kv-components/vue/KvButton.vue
mcstover: Kinda seems like Ghost "is" the new version of Text link...

Otherwise I think the discussion around a text link really comes down to the type style and the padding of the element...

eddieferrer:
@ryan-ludwig There's still a text-decoration underline for secondary buttons with hrefs. Not sure if this is intentional

mcstover (approved):

BoulderBrains (approved):

mcstover (approved):

eddieferrer (approved):

PR #40, opened 2021-05-13
VUE-591: KvMaterialIcon

ryan-ludwig:


ryan-ludwig:
An unfortunate side-effect I've seen from pulling in vue-material-design-icons is that our initial storybook build takes wayyy longer. You can see it running through the 100s of .vue files inside that node_module.

This also has me concerned that we'll be including all of the icons in our UI bundle because of how our vendors bundle works right now. I think doing some webpack bundle analyzing after using this component in UI is a must.

mcstover:
@ryan-ludwig I'd like to spend a little time experimenting with Webpack configs and potentially other base vue packages before we merge this in.

kv-components/vue/stories/KvMaterialIcon.stories.js
```diff
@@ -0,0 +1,155 @@
...
+
+export const Default = DefaultTemplate.bind({});
+Default.args = {
+	name: 'Cart',
+};
+Default.argTypes = {
+	name: {
+		control: {
+			type: 'select',
+			options: getFilenames(require.context('vue-material-design-icons/', true, /\.vue$/)),
```
mcstover, line 41: @ryan-ludwig I'm pretty sure this is why you were getting the longer compiles for storybook. Which... may or may not be ok for Storybook. Maybe we should instead only show a subset instead of the full list available from mdi.

This wouldn't be part of the ui app though b/c we'd only dynamically import what need in context.

mcstover: Actually, I removed this and event without intentionally looping through ALL icons they are all still processed leading to the slowed down builds and performance of storybook. So, I think it's either internal to the imported component or a result of using the computed based dynamic import method...

PR #41, opened 2021-05-13
VUE-591: KvMaterialIcon, alternate version

ryan-ludwig:
This version imports SVG path data from a material icons library. You pass then pass the path data as a prop to KvMaterialIcon component which renders the SVG.

This is in contrast to #40 which imports SFC Vue Components.

The way you set color and size is the same (tailwind classes).

### DIFF
--- a/kv-components/package.json
+++ b/kv-components/package.json
@@ -30,6 +30,7 @@
     "lint": "eslint --ext .js,.vue ./"
   },
   "dependencies": {
+    "@mdi/js": "^5.9.55",
     "kv-tokens": "^0.0.1",
     "vue": "^2.6.12"
   }
--- a/kv-components/vue/KvMaterialIcon.vue
+++ b/kv-components/vue/KvMaterialIcon.vue
@@ -1,0 +1,31 @@
+<template>
+	<span
+		class="inline-flex"
+		aria-hidden="true"
+		role="img"
+	>
+		<svg
+			class="w-full h-full fill-current"
+			width="24"
+			height="24"
+			viewBox="0 0 24 24"
+		>
+			<path :d="icon" />
+		</svg>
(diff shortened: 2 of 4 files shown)


ryan-ludwig:
I think this PR is much more likely to be tree shaken compared to #40. However, I'm not convinced #40 won't be tree-shaken.

This version does spin up storybook much faster.

For either of these PRs we'll need to do webpack bundle analyzing on UI to make sure we're not loading up the 100s of icons in Material Design in a production build.

eddieferrer:
I agree with you, I think this will get tree shaken and is a more efficient approach than #40 

Can we find a way to add the control dropdown back to the story, so that via storybook you can see what all the icons look like? I'm not sure there is a programatic way to do this, but maybe just a file? Some ideas here: https://github.com/google/material-design-icons/issues/241

mcstover:
@eddieferrer @ryan-ludwig There are over 5k icons in mdi so I think we probably just need to continue with the approach of building out the "Common" story with icons we see showing up in our designs. Maybe add a link in this story somewhere out to the mdi visual index.

Otherwise, how are you all feeling about this approach which requires a little more work in that the specific icon needs to be imported before use? I don't really mind that and like the more specific usage and implementation.

mcstover (approved):

eddieferrer (approved):

eddieferrer:
I'm fine with this approach 👍

BoulderBrains (approved):

ryan-ludwig:
@eddieferrer I made a story that looped and displayed all the exported icons from the @mdi/js package. My browser nearly gave up. Probably best to browse and filter from https://materialdesignicons.com/ (though I don't know how they pulled it off)

PR #42, opened 2021-05-17
DIY Tailwind config viewer for our primitives

ryan-ludwig:
A story that builds a visual representation of our tailwind classes by reading in the tailwind config


eddieferrer (approved):
It's kind of weird how you always have to scroll sideways for the xl breakpoint at the bottom because of the max-width on .container.

mcstover (approved):

PR #48, opened 2021-05-19
feat(KvLoadingSpinner): KvLoadingSpinner component

eddieferrer:
Refs: VUE-580


kv-components/vue/KvLoadingSpinner.vue
```diff
@@ -0,0 +1,95 @@
...
+			default: 'medium',
+			validator(value) {
+				return ['small', 'medium', 'large'].includes(value);
+			},
+		},
+		/**
+		 * The color of the loading spinner.
+		 * 'brand', 'white', or 'black'
+		 * */
+		color: {
```
eddieferrer, line 60: Not sure if we want this to be a prop. Alternatively, you can just add the text-* class to the component. Although it does seem nice to limit the available options of colors to just these 3.

kv-components/vue/KvButton.vue
```diff
@@ -99,6 +106,16 @@ export default {
 		return {};
 	},
 	computed: {
+		loadingColor() {
```
eddieferrer, line 109: This is what I inferred from the figma, where each button variant has a specific loading color? Not sure if this is correct, also ghost button doesnt have a loading icon specified in the figma, so I just assumed it was brand color.

mcstover: Seems like a good path to take. Maybe we can call this out to design when they review in storybook.

kv-components/vue/KvLoadingSpinner.vue
```diff
@@ -0,0 +1,95 @@
+<template>
+	<svg
+		class="animate-spin"
+		:class="`${cssClassSize} ${cssClassColor}`"
```
ryan-ludwig, line 4: I don't think this will get picked up by purgeCss. https://tailwindcss.com/docs/optimizing-for-production#writing-purgeable-html

I believe purge only checks the `<template>` area for css class names.

KvButton shows a way to do this all in the template.

ryan-ludwig: It'd be good to verify that though. We'd need to set up purge on a different repo, import this component, and run a prod build, then look at the output CSS file.

kv-components/vue/KvLoadingSpinner.vue
ryan-ludwig: I think the prop and limiting these is good right now 👍

kv-components/vue/KvLoadingSpinner.vue
```diff
@@ -0,0 +1,82 @@
...
+			clip-path="url(#clip0)"
+			stroke-width="2"
+		>
+			<circle
+				class="opacity-low"
+				cx="12"
+				cy="12"
+				r="8"
+				transform="rotate(135 12 12)"
+				stroke="#C4C4C4"
```
ryan-ludwig, line 32: To keep from hard-coding #C4C4C4, does `class="stroke-current text-gray-300 opacity-low` work?
Stolen from https://tailwindcss.com/docs/stroke#usage

ryan-ludwig:
Looks great! Once chromatic is up hopefully we can get this reviewed by Design

kv-components/vue/KvLoadingSpinner.vue
```diff
@@ -0,0 +1,81 @@
...
+		<g
+			clip-path="url(#clip0)"
+			stroke-width="2"
+		>
+			<circle
+				class="stroke-current text-gray-300 opacity-low"
+				cx="12"
+				cy="12"
+				r="8"
+				transform="rotate(135 12 12)"
```
mcstover, line 31: @eddieferrer did you get the rotation spec with the easing and timings from @itseunsoo or Nathan? I'm not sure a constant rotation is going to cover the design they specced out.

eddieferrer: I did not. I actually have not seen any animation from design. This is based on the Interim styleguide figma. Me, @ryan-ludwig, @itseunsoo had a short meeting to discuss the loader in general.

PR #51, opened 2021-05-19
Vue 592

BoulderBrains:
[vue-592](https://kiva.atlassian.net/browse/VUE-592) KvSelect Component

- Created KvSelect component with Material design chevron
- Added KvSelect to storybook with a few variations
     - default
     - disabled
     - labelHidden
     - widthSet


kv-components/vue/stories/KvSelect.stories.js
```diff
@@ -0,0 +1,112 @@
...
+import KvMaterialIcon from '../KvMaterialIcon.vue';
+
+export default {
+	title: 'KvSelect',
+	component: KvSelect,
+};
+
+const Template = (args, { argTypes }) => ({
+	props: Object.keys(argTypes),
+	components: { KvSelect, KvMaterialIcon },
```
ryan-ludwig, line 12: I don't believe KvMaterialIcon is used anymore

kv-components/vue/KvSelect.vue
```diff
@@ -0,0 +1,72 @@
+<template>
+	<div>
+		<span class="relative">
```
ryan-ludwig, line 3: Small thing, but I think you can remove this span by putting the "relative" on the div above

mcstover:
@BoulderBrains could you please squash all your commits into 1 commit that uses the conventional commits spec?
https://www.conventionalcommits.org/en/v1.0.0/

I think this would be a `feat` type.

kv-components/vue/KvSelect.vue
emuvente: +1

BoulderBrains: Wasn't able to remove the double container, but reworked it as we discussed @ryan-ludwig

ryan-ludwig: @emuvente An issue came up with conflicting classes for positioning. If you were to use KvSelect somewhere with
`<kv-select class="absolute top-0 right-0">` 
that absolute class would create a conflict if KvSelect looked like
```
<template>
  <div class="relative...">
    <select...
 ```

Keeping the outer div fairly class-free seems like something we should be mindful of for these Kv* components.

File: package-lock.json [hunk 3]
```diff
@@ -32,6 +32,7 @@
       "dependencies": {
         "@mdi/js": "^5.9.55",
         "kv-tokens": "^0.0.1",
+        "tailwindcss": "npm:@tailwindcss/postcss7-compat@^2.1.0",
```
ryan-ludwig, line 35: Is there a change to package.json that needs to be committed?

ryan-ludwig:
This is looking great @BoulderBrains !

package-lock.json
BoulderBrains: I've don't have any package.json changes, this was added when you and I were on zoom earlier today. Without this line lint throws this error and prevents me from committing my code: 

/Users/joshua/kiva/kv-ui-elements/kv-tokens/configs/tailwind.config.js
  1:24  error  'tailwindcss' should be listed in the project's dependencies. Run 'npm i -S tailwindcss' to add it  import/no-extraneous-dependencies

✖ 1 problem (1 error, 0 warnings)

kv-components/vue/stories/KvSelect.stories.js
```diff
@@ -0,0 +1,107 @@
...
+				Choose an example
+			</label>
+			<kv-select
+				id="example-id"
+				:disabled="disabled"
+				@change="onChange"
+				v-model="exampleModel"
+				class="mt-2"
+			>
+					<option class="px-3 py-2" value="example1">Example one</option>
```
eddieferrer, line 29: The classes on the options are kind of annoying Because when using KvSelect, you'll have to remember to add these classes to your options elements. Could we possibly add these classes as mixins in the css of KvSelect? I'm thinking something like:
```
option {
  @apply px-3 py-2;
}

File: kv-components/vue/KvSelect.vue [hunk 5]
```diff
@@ -0,0 +1,72 @@
...
+			default: false,
+		},
+	},
+	data() {
+		return {
+			mdiChevronUp,
+		};
+	},
+	methods: {
+		onChange(event) {
```
eddieferrer, line 66: Is onChange needed? We already have a v-model that changes and contains this value, so anything that could be done with the onchange event could also be done with watching the v-model.

kv-components/vue/stories/KvSelect.stories.js
eddieferrer: Actually I think the `px-3 py-2` classes can be removed completely

kv-components/vue/KvSelect.vue
```diff
@@ -0,0 +1,72 @@
...
+<script>
+import { mdiChevronUp } from '@mdi/js';
+import KvMaterialIcon from './KvMaterialIcon.vue';
+
+export default {
+	components: {
+		KvMaterialIcon,
+	},
+	// v-model will change when select value updates
+	model: {
```
eddieferrer, line 32: This block is unnecessary

eddieferrer:
Alternatively, instead of using a default slot for all the options, we could pass in a prop that is an array of options...
`:options='options'`

options = [
{ label: 'example1', value: 'example1value' }
]

eddieferrer:
On smaller resolutions, when the select doesnt fit, the icon wraps by itself: 
Screen Shot 2021-05-20 at 8 11 55 AM

kv-components/vue/stories/KvSelect.stories.js
BoulderBrains: Thanks for catching this @eddieferrer, this was left over from me trying to style the open state of the dropdown before I realized that we're going with browser default and not what's shown in figma file.

kv-components/vue/KvSelect.vue
ryan-ludwig: @eddieferrer This is needed to get v-model working. Without it, `exampleModel` does not update when the select is changed in the example below.
```
	data: () => ({
		exampleModel: 'example2',
	}),
	template: `
		<div>
			exampleModel = {{exampleModel}}
			<kv-select
				v-model="exampleModel"
			>
					<option value="example1">Example one</option>
					<option value="example2">Example two</option>
					<option value="example3">Example three</option>
			</kv-select>
		</div>`,

Comment (eddieferrer) [reply kv-components/vue/KvSelect.vue:32]: You're totally right, I thought this was the default, but I think the default is input and not change.

**Comment by BoulderBrains** [comment]
@eddieferrer & @ryan-ludwig 
I've done some testing locally and I was unable to reproduce the label breaking on small screens. If the user sets the width of the select on the highest level div container, ie the one with inline-flex on it, the label goes behind the icon gracefully on small screens without applying the tailwinds class max-w-full (aka max-width: 100%). Am I misunderstanding the scenario that you're describing? 
width

**Comment by eddieferrer** [comment]
> @eddieferrer & @ryan-ludwig
> I've done some testing locally and I was unable to reproduce the label breaking on small screens. If the user sets the width of the select on the highest level div container, ie the one with inline-flex on it, the label goes behind the icon gracefully on small screens without applying the tailwinds class max-w-full (aka max-width: 100%). Am I misunderstanding the scenario that you're describing?
> width

I can't reproduce it now either. 👍  I think the wrapping div change fixed it

**Comment by BoulderBrains** [comment]
> Alternatively, instead of using a default slot for all the options, we could pass in a prop that is an array of options...
> `:options='options'`
> 
> options = [
> { label: 'example1', value: 'example1value' }
> ]

The combination of slots being the more flexible option and the current KvSelect component being built with <slot></slot> is why I chose to go with this method. But perhaps because the only thing we will be putting into the slot of a KvSelect would be the options to display, we could replace the <slot> with the :options='options' prop? 

Is that why you would push toward using the prop instead of the slot option @eddieferrer? If not, what's the benefit of changing from slot -> prop?

**Comment by eddieferrer** [comment]
> > Alternatively, instead of using a default slot for all the options, we could pass in a prop that is an array of options...
> > `:options='options'`
> > options = [
> > { label: 'example1', value: 'example1value' }
> > ]
> 
> The combination of slots being the more flexible option and the current KvSelect component being built with is why I chose to go with this method. But perhaps because the only thing we will be putting into the slot of a KvSelect would be the options to display, we could replace the with the :options='options' prop?
> 
> Is that why you would push toward using the prop instead of the slot option @eddieferrer? If not, what's the benefit of changing from slot -> prop?

I don't have a strong preference either way. 

While the slots is more flexible, it also generates more code in the parent component and a lot of times I think you'll be iterating through some array to generate the option tags anyway

**Review by mcstover — APPROVED** [review]
This is a good first version.

We might want to create a spike ticket around going the props route that can get into a deeper discussion about things like dividers and multi select that might eventually be needed in this component.

**Comment by eddieferrer** [comment]
@BoulderBrains Please remove or checkout from main kv-tokens/configs/tailwind.config.js  and package-lock so they aren't included in this

**Review by emuvente — APPROVED** [review]
👍 looks good

**Comment by BoulderBrains** [comment]
> @BoulderBrains Please remove or checkout from main kv-tokens/configs/tailwind.config.js and package-lock so they aren't included in this

The only changed files are now KvSelect.vue and KvSelect.stories.js.

File: kv-components/vue/KvSelect.vue [hunk 7]
```diff
@@ -0,0 +1,72 @@
+<template>
+	<div class="inline-flex">
+		<div class="relative w-full">
+			<select
+				:id="id"
+				:disabled="disabled"
+				:value="value"
+				class="text-base h-6 pr-4 pl-1 border border-gray-300 rounded-sm appearance-none w-full"
```
ryan-ludwig, line 8: I'd suggest `.pl-2` instead of `.pl-1` here to match their Figma doc.

kv-components/vue/KvSelect.vue
```diff
@@ -0,0 +1,72 @@
...
+		},
+	},
+	data() {
+		return {
+			mdiChevronDown,
+		};
+	},
+	methods: {
+		onChange(event) {
+			// emit a vue event
```
ryan-ludwig, line 67: Please remove this comment and add a JsDoc here of 
```
			/**
			 * The value that the select has changed to
			 * @event change
			 * @type {Event}
			 */			 
```

ryan-ludwig (approved):
Two minor tweaks, overall looks great!

eddieferrer (approved):
Nice!

BoulderBrains:
This branch got really messy when trying to squash and reword commits because I merged MAIN into it so many times. I'm closing this PR, checking out main, starting a fresh branch and picking out my changes more cleanly, not an ideal solution, but things are really tangled up currently.

PR #49, opened 2021-05-19
VUE-521 VUE-572 Release process

emuvente:
This prepares the repo for publishing releases to NPM using Lerna. I will do the initial publish manually once this is in main, and then make a follow-up PR for automated releases.

PR #53, opened 2021-05-20
fix(build): Move scoped workspaces to directory matching scope

emuvente:
The changes introduced in #49 cause an invalid package-lock.json to be created when `npm install` is run. After that, any calls to `npm ci` result in this error:
```
npm ERR! The `npm ci` command can only install with an existing package-lock.json or
npm ERR! npm-shrinkwrap.json with lockfileVersion >= 1. Run an install with npm@5 or
npm ERR! later to generate a package-lock.json file, then try again.
```
This error was caused by the package names not precisely matching the workspace directory. The `@kiva/kv-components` package was in the `kv-components/` directory when it should have been in the `@kiva/kv-components/` directory (`kv-tokens` was incorrect in the same way).

This PR fixes the issue by moving the `kv-components/` and `kv-tokens/` directories into the `@kiva/` directory. For the other PRs that are open right now that make changes to files in those directories, those can be merged first and then I'll resolve any conflicts in this PR.

An alternative solution would be to not scope those packages, though that would mean having to ensure those package names didn't already exist in NPM, or not publishing to NPM at all.


mcstover (approved):

eddieferrer (approved):

ryan-ludwig (approved):

BoulderBrains (approved):
👍

PR #55, opened 2021-05-21
build: Temporarily use chromaui/action-canary@v1 to fix broken storyb…

mcstover:
…ook deployment.


emuvente (approved):

eddieferrer:
I think this is unnecessary, chromatic fixed it on their end late yesterday

PR #57, opened 2021-05-25
 feat(KvContentfulImg): Component that renders image from contentful

eddieferrer:
Some differences/enhancements from the component currently used in UI

* Added some more prop validators.
* Split crop prop to fit / focus props. This will allow a more direct mapping of the props to the contentful image api.


mcstover (approved):

emuvente (approved):
👍

PR #59, opened 2021-06-03
VUE-604 letter spacing 

ryan-ludwig:
Convert letter-spacing of h1,h2,h3,h4s to ems so the existing UI letter spacing will look fairly normal.


ryan-ludwig:
It's odd there's a visual diff on this. Looking into it.

ryan-ludwig:
Oh, thanks visual differ!

mcstover (approved):
It doesn't seem like this materially changes the end visual result vs rems as it's only applied to letterSpacing. So we should be ok.

PR #60, opened 2021-06-08
Add ability to set custom image sources across source sets. This enab…

mcstover:
…les an adaptive image setup for using different images across viewport sizes.


emuvente (approved):
Nice!

ryan-ludwig (approved):
Sweet!

PR #62, opened 2021-06-09
fix: adding tailwinds override for foundation classes applied globall…

BoulderBrains:
…y in /ui

Within the /kv-ui-elements - KvSelect component have the tailwinds class of 'appearance-none', which hides the browser default < select > appearance and allows us to the use material design chevron that we're using. That worked well in that repo, check it out: 
Screen Shot 2021-06-09 at 2 32 29 PM

Once we pulled the kv-ui-elements components into UI, we have foundation classes applied globally which puts a background image in the < select > and results in this state: 
Screen Shot 2021-06-09 at 2 27 15 PM

The additional class I've added to the KvSelect will override the foundation background image that's globally applied to < select > elements.

### DIFF
--- a/@kiva/kv-components/vue/KvSelect.vue
+++ b/@kiva/kv-components/vue/KvSelect.vue
@@ -6,7 +6,7 @@
 				:id="id"
 				:disabled="disabled"
 				:value="value"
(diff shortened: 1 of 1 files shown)


emuvente (approved):
:+1:

mlibazisi (approved):
👍

ryan-ludwig (approved):

PR #64, opened 2021-06-15
Background image adjustments for KvContentfulImg

emuvente:


@kiva/kv-components/vue/KvContentfulImg.vue
```diff
@@ -86,7 +91,6 @@ export default {
 		* */
 		fallbackFormat: {
 			type: String,
-			required: true,
```
emuvente, line 89: Since this prop has a default it doesn't need to be required

mcstover: Is that field needed to activate the validator?

@kiva/kv-components/vue/KvContentfulImg.vue
```diff
@@ -45,6 +45,11 @@
 			<!-- browser doesn't support webp or browser doesn't support picture element -->
 			<img
 				class="tw-max-w-full tw-max-h-full"
+				style="
+					width: inherit;
+					height: inherit;
+					object-fit: inherit;
```
mcstover, line 51: guessing we don't have tw classes for these.

@kiva/kv-components/vue/KvContentfulImg.vue
mcstover: no it doesn't appear to be.

mcstover (approved):

@kiva/kv-tokens/configs/tailwind.config.js
```diff
@@ -124,6 +124,10 @@ module.exports = {
 		}),
 		extend: {
 			typography: kivaTypography.proseOverrides, // prose plugin overrides
+			height: {
+				40: rem(320),
```
ryan-ludwig, line 128: I think this works well for now. If we end up pulling in the aspect-ratio plugin later on, it'd be nice to use that and keep these misc. height values out of the config.

ryan-ludwig (approved):

@kiva/kv-components/vue/KvContentfulImg.vue
emuvente: None that I could find

PR #66, opened 2021-06-16
VUE-611 Grid and container components

ryan-ludwig:
Two components that can be used separately or together. 
KvPageContainer constrains and centers content per the design spec.
KvGrid is a light wrapper around tailwind's grid class that keeps the column gaps to the design specification.


mcstover (approved):
Makes sense and provides a nice centralized place to control our approach to the "container".

emuvente (approved):
👍

ryan-ludwig:
Going to merge this so it can be used while we wait for design feedback since the component APIs will not change.

BoulderBrains:
👍

PR #68, opened 2021-06-17
VUE-583 KvProgressBar improvements

ryan-ludwig:
fix(KvProgressBar): Add ariaLabel prop, fix aria attribute DOM placement, add min prop in case it's non-zero.
perf(KvProgressBar): Animate using css transforms instead of width


emuvente (approved):
👍

mcstover (approved):

BoulderBrains:
👍

PR #72, opened 2021-06-23
feat(KvTabs): Accessible components for implementing a tab UI pattern

ryan-ludwig:
VUE-594

### DIFF
--- a/@kiva/kv-components/vue/KvTab.vue
+++ b/@kiva/kv-components/vue/KvTab.vue
@@ -1,0 +1,55 @@
+<template>
+	<button
+		:id="`kv-tab-${this.for}`"
+		class="tw-text-h3 tw-mb-1.5 tw-whitespace-nowrap"
+		:class="{ 'tw-text-gray-300  hover:tw-text-action-700' : !isActive }"
+		role="tab"
+		:aria-selected="isActive"
+		:aria-controls="`kv-tab-panel-${this.for}`"
+		:tabindex="isActive ? null : -1"
+		@click="handleTabClicked"
+	>
+		<slot></slot>
+	</button>
+</template>
+
+<script>
+export default {
+	inject: ['$KvTabContext'],
+	props: {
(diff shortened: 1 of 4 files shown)


ryan-ludwig:
Any feedback on how you imagine composing these tabs would be great. Is the slot syntax working for you? Does the provide/inject pattern make sense to you here?

@kiva/kv-components/vue/KvTab.vue
```diff
@@ -0,0 +1,55 @@
...
+			const { navItems, selectedIndex } = this.$KvTabContext;
+			return navItems[selectedIndex]?.for === this.for;
+		},
+		index() {
+			const { navItems } = this.$KvTabContext;
+			return navItems?.findIndex((navItem) => navItem.for === this.for);
+		},
+	},
+	mounted() {
+		this.$KvTabContext.navItems.push(this);
```
ryan-ludwig, line 47: This felt a little funny at first, but seems to work OK. We could also populate the navItems array inside KvTabs.vue by checking `$slots.tabNav` if this seems too weird.

mcstover: I think this seems ok for now. My initial concern was just the extraneous data associated with each instance of a KvTab. This is a fairly light component...In other more complex cases I would def recommend we inject only what is needed.

mcstover (approved):

emuvente:
The alternatives to provide/inject that come to mind would involve more complex code, so this is fine by me! I was initially concerned that it wouldn't work if you had multiple KvTabs on the page, but then I saw the story with multiple KvTabs working just fine. This looks great!

emuvente (approved):

mcstover:
👍

BoulderBrains (approved):
The underline animation growing and shrinking when transition between tabs looks really nice.

mlibazisi (approved):
👍

PR #76, opened 2021-07-01
VUE-587 KvLightbox

ryan-ludwig:


@kiva/kv-components/vue/KvLightbox.vue
```diff
@@ -0,0 +1,285 @@
...
+			type: Boolean,
+			default: false,
+		},
+		/**
+		 * Appearance and role of the lightbox
+		 * @values lightbox, alert
+		 * */
+		variant: {
+			type: String,
+			default: 'lightbox',
```
emuvente, line 186: You could use a validator function for this to ensure correct usage:
```
validator(value) {
	 return ['alert', 'lightbox'].indexOf(value) > -1;
}
```

@kiva/kv-components/vue/KvLightbox.vue
```diff
@@ -0,0 +1,285 @@
...
+		 * */
+		preventClose: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	data() {
+		return {
+			mdiClose,
+			hideOthers() {}, // reference to aria-hide function
```
emuvente, line 208: I don't see `hideOthers` used anywhere. Is this supposed to be `makePageInert`?

emuvente (approved):
This looks awesome! There's a lot of great things contained here, good work!

@kiva/kv-components/vue/KvLightbox.vue
ryan-ludwig: Yes, missed it in a refactor. Thanks!

mcstover (approved):

@kiva/kv-components/vue/KvLightbox.vue
```diff
@@ -0,0 +1,288 @@
...
+ * Alert dialog - https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role
+ *
+ * - [x] focus moves to the first non-destructive control, rather than the close button
+ * - [x] role = alertDialog
+ * - [x] aria-describedby is set to the id of the dialog body
+ *
+ * Printing
+ *
+ * - [x] Only prints the contents of the lightbox when open
+ * TODO: ensure purge doesn't remove these classes
```
ryan-ludwig, line 160: @emuvente, @mcstover   Any ideas on how to keep these classes from being purged? They get added by a .js util.  I tried adding them to a `safelist` in the shared tailwind config purge object, but that will get overridden by the project tailwind config. As in, you can't "extend" the purge part of the config.

One way would be a blank div like `<div class="tw-hidden print:tw-invisible print:tw-visible..."></div>` in this file's template. Pragmatic but not super clean.

emuvente: When including kv-components in other projects, we already need to ensure that the vue files from kv-components are included in the purge options, so we could include directions in the README about which purge options to be sure to use when adding kv-components to a project.

Your suggestion of an invisible div with the classes is not the worst alternative though.

mcstover: Reviewing the docs it seems like the syntax changes when introducing more specific purge configuration. For example when setting up "safeList" options they live alongside a "content" array. We currently don't purge anything (and have not specified which files to include/exclude) in kv-components so adding the safeList without corresponding configs just purged everything except the print styles.

Thus, I think we could add the safeList for print styles in the UI repo as Shua mentioned. We might also need to change our syntax there to have the "contents" array alongside the "safeList" array instead of just a root array in our tailwind config.

mlibazisi (approved):

BoulderBrains (approved):
👏

@kiva/kv-components/vue/KvLightbox.vue
ryan-ludwig: I spun up a simple Vue cli app + tailwind  and did some various tests with purge. It's very lenient. Classes written in the `<script>` tag of .vue files get preserved, like 
```
addAClass() { this.classList.add('tw-float-left') }
```
Commented out classes get picked up, like
```
<!-- <div class="tw-float-right"></div> -->
```
 Basically if it finds any possible tailwind class anywhere in any file it seems to preserve it from being purged.

ryan-ludwig: For this PR we're adding tailwind classes from `utils/printing.js` and `utils/scrollLock.js`, which won't get pulled in by the current purge settings on UI. So I think we just need to add `'./node_modules/@kiva/kv-components/utils/**/*.js'` to the list on UI and we're set.

I'll make a separate PR explaining this in the README.

PR #80, opened 2021-07-09
Update Tailwind, use JIT

ryan-ludwig:
I've been running into some issues matching designs when arbitrary widths are used (like 550px) inside of a media query. Since it's inside a media query, inline styles can't be used to fudge it.

Since v2.1 Tailwind introduced JIT mode which among other things,
> Generate arbitrary styles without writing custom CSS. Ever needed some ultra-specific value that wasn’t part of your design system, like top: -113px for a quirky background image? Since styles are generated on demand, you can just generate a utility for this as needed using square bracket notation like top-[-113px]. Works with variants too, like md:top-[-113px].

https://tailwindcss.com/docs/just-in-time-mode

The other thing I like a lot is the purge settings are the same in dev or prod mode. So if something's broken in dev because of purge you'll see it there first.

It does still say JIT mode is in "preview", thought they've released a minor and several patches since. I didn't notice any issues on here or on UI.

### DIFF
--- a/@kiva/kv-components/package.json
+++ b/@kiva/kv-components/package.json
@@ -15,7 +15,6 @@
     "@storybook/addon-postcss": "^2.0.0",
     "@storybook/addon-storysource": "^6.3.2",
     "@storybook/vue": "^6.3.2",
-    "@tailwindcss/postcss7-compat": "^2.1.0",
     "autoprefixer": "^9.8.6",
     "babel-loader": "^8.2.2",
     "chromatic": "^5.9.2",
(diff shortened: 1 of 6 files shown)


ryan-ludwig:
Interesting, since I'm dynamically applying a lot of classnames in the primitives viewer page, purge isn't picking those up and so it discards them. This is causing the "primitives" visual diff on chromatic.

emuvente (approved):
Nice!

mcstover (approved):
How does this in turn affect the UI (or other app usage) of this more dynamic tailwind.config fire? Will it be expected that we need to add things to this safelist periodically?

ryan-ludwig:
> How does this in turn affect the UI (or other app usage) of this more dynamic tailwind.config fire? Will it be expected that we need to add things to this safelist periodically?

@mcstover Each repo is in control of their own purge settings, since purge doesn't get extended like the theme does. The safelist changes here are just so our storybook shows classes that aren't in use in this repo currently. 

Consumers of this repo won't have purge changes to update unless they are doing something special.

PR #81, opened 2021-07-12
VUE-589 KvToast

ryan-ludwig:
This should be nearly a drop-in replacement for KvTipMessage, though I'd like to handle watching the $route within TheTipMessage.vue instead.

### DIFF
--- a/@kiva/kv-components/vue/stories/KvToast.stories.js
+++ b/@kiva/kv-components/vue/stories/KvToast.stories.js
@@ -1,0 +1,74 @@
+import KvButton from '../KvButton.vue';
+import KvToast from '../KvToast.vue';
+
+export default {
+	title: 'KvToast',
+	component: KvToast,
+	args: {
+		message: 'Sucessfully added to basket!',
+		persist: false,
+		type: 'confirmation',
+	},
+	argTypes: {
+		message: {
+			control: {
+				type: 'text',
+			},
+		},
+		type: {
+			options: ['confirmation', 'warning', 'error'],
+			control: {
+				type: 'select',
+			},
+		},
+		persist: {
+			control: {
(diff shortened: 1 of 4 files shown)


BoulderBrains:
This toast message looks really nice, I like the type warning on the left (indicated by color & icon) and the messaging along the right. The entrance and exit animations are nice and smooth 👍 

The only issue I found was that for med->large screens, in the "With long text and html" story, the messaging wasn't contained to the view completely. 
Screen Shot 2021-07-13 at 8 44 29 AM

mcstover (approved):

BoulderBrains (approved):

@kiva/kv-components/vue/stories/KvToast.stories.js
```diff
@@ -0,0 +1,74 @@
...
+		KvToast,
+		KvButton,
+	},
+	template: `
+		<div>
+			<kv-button @click="showToast(message, type, persist)">Show Toast</kv-button>
+
+			<!-- div below is a kludge for storybook docs -->
+			<div class="tw-fixed tw-inset-0 tw-pointer-events-none">
+				<div class="tw-fixed tw-z-toast tw-left-0 tw-right-0 tw-top-2 tw-pointer-events-auto">
```
mcstover, line 46: Should we have these style classes (ie tw-fixed, tw-z-toast and positiontion classes) embedded in the toast component? I think with those added, this component becomes something that can be used as a stand along component in some cases so graphql isn't a dependency.

ryan-ludwig: Since the toast may change positions depending on breakpoint and the height of the navigation, planning on handling positioning at the app level.

PR #85, opened 2021-07-22
Setup Jest and Test Library

ryan-ludwig:
I've been reading more about component testing lately and would like to add some more to this component library. We currently have snapshot testing and visual regression testing handled by Chromatic. The Jest work is mainly to test the component APIs themselves and I'd like to also hook up some automated A11y checks in the near future.

Ran into a regression on KvButton where we lost the ripple animation which wasn't caught by Chromatic, and I've got some form elements in progress that have a lot of events and methods to test (blur, focus, change, etc.) so it seemed like a good time to get this going.

I'm pretty new to testing! Very open to any feedback on what to test and how to write them better.


@kiva/kv-components/tests/unit/specs/components/KvButton.spec.js
```diff
@@ -0,0 +1,35 @@
+import { mount } from '@vue/test-utils';
+import KvButton from '../../../../vue/KvButton.vue';
+import { VariantPrimary, WithHref, StateLoading } from '../../../../vue/stories/KvButton.stories';
```
mcstover, line 3: I "think" the built in webpack aliases should work here or maybe we'll need to set them up elsewhere.

ryan-ludwig:
Oh interesting, seeing now that the official Vue documentation points towards using [Vue Testing Library](https://testing-library.com/docs/vue-testing-library/intro/) instead of Vue Test Utils directly - https://vuejs.org/v2/guide/testing.html#Recommendations

mcstover:
@ryan-ludwig @emuvente I think using the "Testing Library" overlay would be worth a try in this contrained world of view components. It uses vue-test-utils underneath AND has a Jest Dom plugin too.

emuvente (approved):
This looks great!

mcstover (approved):
Great start to getting code test infra in place for this repo!

eddieferrer (approved):

PR #91, opened 2021-07-30
VUE-643: Try adding sideEffects:false to improve tree shaking

ryan-ludwig:


eddieferrer (approved):
Worth a try

ryan-ludwig:
Made no difference that I could tell

PR #93, opened 2021-08-02
VUE-609 KvCheckbox

ryan-ludwig:


@kiva/kv-components/vue/KvSwitch.vue
```diff
@@ -78,14 +80,7 @@ export default {
 			default: false,
 		},
 		/**
-		 * Value of the input if this component is used within a form submission
```
ryan-ludwig, line 81: Removing `value` prop from checkbox and switch since I don't think it would ever be used in a Vue environment.

@kiva/kv-components/vue/KvCheckbox.vue
```diff
@@ -0,0 +1,116 @@
+<template>
+	<div>
```
emuvente, line 2: Since this div has only one child element, you could drop this div and use the label as the root element for this component

ryan-ludwig: I'd like to keep this wrapping div to keep things predicitable, since the label has a display class, `tw-inline-flex` on it. This could cause issues if you were trying to show/hide things responsively like
`<kv-checkbox class="sm:tw-hidden lg:tw-block" />`

@kiva/kv-components/vue/KvCheckbox.vue
```diff
@@ -0,0 +1,116 @@
...
+
+<script>
+import { nanoid } from 'nanoid';
+
+/**
+ * Use as you would an <input type="checkbox" />
+ * Does not handle 'indeterminate' state at this time.
+*/
+
+export default {
```
emuvente, line 66: since you're applying $attrs and $listeners to the input component, I think you'll also want to set `inheritAttrs: false` in the component options so that the attributes and listeners don't get applied to both the root div and the input element https://vuejs.org/v2/api/#inheritAttrs

emuvente (approved):
One comment, but otherwise looks great!

BoulderBrains (approved):

@kiva/kv-components/vue/KvCheckbox.vue
ryan-ludwig: thank you!

PR #95, opened 2021-08-04
fix(KvTabs): Prevent different height tab contents from overflowing t…

ryan-ludwig:
…heir container during a transition


ryan-ludwig:
There's kindof an ugly jump when switching tabs on borrower profile when they have different heights. You notice it because of the background colors. This PR stops the content from fading out before fading in the new content, since the height of the container changes immediately.

emuvente (approved):
:+1:

PR #101, opened 2021-08-18
fix:ensure inherited styles in KvContentfulImg by removing non-breaki…

mcstover:
…ng spaces from style string.


mcstover:
Styles were not being inherited and were showing as invalid in the web inspector.

emuvente (approved):

eddieferrer (approved):
Good catch! I think it has to be in one line like this fix, or 
```
style="`
```

mcstover:
Line length is at 78 so clear of linter

PR #102, opened 2021-08-19
VUE-646 color theming

ryan-ludwig:
* Changes from using Tailwind's default color system to using themable colors using CSS custom properties
* Updates references throughout this repo to use new themable class names
* Adds `kv-theme-provider` component which is used to apply a theme to child elements
* Adds storybook "dark mode" addon to test that theming is working on any story
* Creates "default", "dark", and "mint" themes in primitives
* Breaks Styleguide.stories.js into multiple files

Sorry for the giant PR!

I'd like to improve the documentation, and possibly change how to pass a theme to KvThemeProvider. 
Probably more important though is to get something we're mostly happy with in order to get tailwind classnames changed in UI. Then make docs and other improvements after.


eddieferrer (approved):
Looks good. I could go either way on the theme wrapper component. If we were to use the theme wrapper as suggested by @emuvente I think we'll still be able to use `component: is` to dynamically load the correct wrapper for the story cards.

ryan-ludwig:
Thanks for taking a look everyone. Will merge this to get UI onto the new class names, then loop back for improving theme imports and better documentation.

PR #104, opened 2021-08-20
feat: modify KvCarousel to allow for responsive slides

eddieferrer:
I had to modify this component to work with the use case of the upcoming cause based landing page.

See story: 
https://608b4cf87f686c00213841b1-qpcaijlqvm.chromatic.com/iframe.html?id=kvcarousel--responsive-slides&args=&viewMode=story

For the use case we are trying to solve.

### DIFF
--- a/@kiva/kv-components/vue/stories/KvCarousel.stories.js
+++ b/@kiva/kv-components/vue/stories/KvCarousel.stories.js
@@ -199,3 +199,22 @@ export const loadingLoanCardExample = () => ({
 		},
 	},
 });
+
+export const ResponsiveSlides = () => ({
+	components: {
+		KvCarousel,
+		KvButton,
+	},
+	template: `
+		<kv-carousel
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/vue/stories/KvCarousel.stories.js
```diff
@@ -199,3 +199,22 @@ export const loadingLoanCardExample = () => ({
 		},
 	},
 });
+
+export const ResponsiveSlides = () => ({
```
eddieferrer, line 203: This is the use case we are trying to solve. We want to display 2 slides at a time, at a fixed width that is approximately 50%.  But when we get to smaller resolutions, we want to display a single slide, that is responsive.

mcstover (approved):
Def a useful improvement!

ryan-ludwig (approved):

PR #117, opened 2021-09-23
fix: Remove unnecessary lightbox class and webkit fix, make lightbox …

eddieferrer:
…stories initialize with open lightboxes for better visual diffs


ryan-ludwig (approved):
Nice!

PR #118, opened 2021-09-23
feat: Add active state to KvButton

ryan-ludwig:
Looks the same as the hover state, but doesn't change when hovered. First time it's shown up in a design that I've seen, but maybe more on the way.

### DIFF
--- a/@kiva/kv-components/vue/KvButton.vue
+++ b/@kiva/kv-components/vue/KvButton.vue
@@ -18,18 +18,7 @@
 			ref="buttonInnerRef"
 			class="tw-inline-flex tw-w-full tw-justify-center tw-items-center tw-rounded
 				tw-min-h-6 tw-relative tw-overflow-hidden tw-border tw-font-medium"
-			:class="{
-				'tw-bg-action hover:tw-bg-action-highlight tw-text-primary-inverse tw-border-action hover:tw-border-action-highlight'
-					: variant === 'primary',
-				'tw-bg-primary hover:tw-bg-secondary tw-text-primary tw-border-tertiary hover:tw-border-primary'
-					: variant === 'secondary',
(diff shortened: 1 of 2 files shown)


eddieferrer (approved):

ryan-ludwig:
Merging so we can get the version bumped in UI.

@kiva/kv-components/vue/KvButton.vue
```diff
@@ -18,18 +18,7 @@
...
-				'tw-bg-primary hover:tw-bg-secondary tw-text-primary tw-border-tertiary hover:tw-border-primary'
-					: variant === 'secondary',
-				'tw-bg-danger hover:tw-bg-danger-highlight tw-text-primary-inverse tw-border-danger hover:tw-border-danger-highlight'
-					: variant === 'danger',
-				'tw-bg-primary-inverse tw-text-primary-inverse tw-border-primary hover:tw-border-secondary'
-					: variant === 'link',
-				'tw-bg-primary hover:tw-bg-secondary tw-text-primary tw-border-transparent'
-					: variant === 'ghost',
-			}"
+			:class="computedClass"
```
mcstover, line 21: Won't this change now mean some of these classes are skipped by the purge? Specifically the border and bg classes may need to be added to the safelist.

ryan-ludwig: @mcstover I previously thought that the Tailwind classes needed to be in the markup. That was incorrect, they just need to be anywhere in the file. As long as we're not doing string interpolation like `tw-bg-${somvar}` we're good. 

I feel confident in this since there were no visual diffs in Chromatic.

mcstover: Wonder if the recent update fixed (or added) that functionality. I think my earlier attempts to do this had issues with purge. Either way, glad this is the case!

PR #119, opened 2021-09-28
VUE-729 disabled carousel btns

ryan-ludwig:


@kiva/kv-components/vue/KvCarousel.vue
```diff
@@ -33,11 +33,12 @@
...
 			<button
-				class="tw-text-gray-800 disabled:tw-text-gray-300
+				class="tw-text-primary
 					tw-rounded-full
-					tw-border-2 tw-border-text-gray-800
+					tw-border-2 tw-border-primary
 					tw-h-4 tw-w-4
-					tw-flex tw-items-center tw-justify-center"
+					tw-flex tw-items-center tw-justify-center
+					disabled:tw-opacity-low disabled:tw-cursor-default"
```
ryan-ludwig, line 41: Switched to low-opacity instead of text color since that's what we do for kv-button. 
Considering just making a change to the global tailwind reset css, like
```
button:disabled {
  opacity: 0.3 (or whatever the theme value is)
  pointer-events: none (will give us a non-hand cursor and prevents the ripple effect  on kv-button
}
```

ryan-ludwig: But that's potentially not what we always want... meh, probably better to keep it in tw classes

eddieferrer (approved):
Looks good. I was thinking about this recently, thanks for getting it done!

mcstover (approved):

PR #121, opened 2021-09-29
fix(KvCheckbox): Allow checkboxes to use an array in v-model

ryan-ludwig:
VUE-715


ryan-ludwig:
Allows for binding multiple checkboxes to the same array, like 
https://vuejs.org/v2/guide/forms.html#Checkbox

mcstover (approved):
A little ;) more code to support watching both but looks right.

BoulderBrains (approved):

eddieferrer (approved):
👍

emuvente (approved):
Nice!

PR #132, opened 2021-11-05
fix(KvLightbox): prevent lightbox references from being null when the…

ryan-ludwig:
… lightbox is closed


ryan-ludwig:
Taken from https://github.com/kiva/ui/commit/9263531b3a58636d675fc0c0060c5abadfb756ee

emuvente (approved):

PR #133, opened 2021-11-08
fix: use conventional commit with lerna for CI

ryan-ludwig:
Changes the default commit message from "Publish" so the linter passes on CI releases.


mcstover (approved):
makes sense, glad there was a way to specify!

PR #134, opened 2021-11-08
feat(KvCheckbox): add a prop and visual styling to tell a user the input is invalid

ryan-ludwig:
Similar to https://608b4cf87f686c00213841b1-jzxgtxmnny.chromatic.com/?path=/story/forms-kvtextinput--invalid


ryan-ludwig:
Oh interesting, I see there's an "error" slot in KvTextInput where error text goes. This might not be the complete solve then.

mcstover (commented):
Seems ok

ryan-ludwig:
Added VUE-818 as a follow-up ticket

mcstover (approved):

emuvente (approved):

PR #135, opened 2021-11-09
fix(KvLightbox): allow for an empty lightbox body

ryan-ludwig:
Useful when you have an alert that's just a lightbox title + 2 buttons

image


mcstover (approved):
Simple enough!

PR #138, opened 2021-11-24
Generate SCSS vars based on tokens files using Style Dictionary tool

ryan-ludwig:


@kiva/kv-tokens/dist/scss/_variables.scss
```diff
@@ -0,0 +1,172 @@
...
+$space-16: 128;
+$space-0-5: 4;
+$space-1-5: 12;
+$space-2-5: 20;
+$space-3-5: 28;
+$space-4-5: 36;
+$fonts-sans: PostGrotesk;
+$font-sizes-base-sm: 16;
+$font-sizes-base-md: 16;
+$font-sizes-base-lg: 17;
```
ryan-ludwig, line 114: All of these unitless values are kindof a pain to deal with in SASS and CSS. I wonder if we should add 'px' to them in the tokens file directly.

@kiva/kv-tokens/dist/css/variables.css
```diff
@@ -0,0 +1,175 @@
+/**
+ * Do not edit directly
```
ryan-ludwig, line 2: Should we include these dist files in our repo, or only include them as part of the npm package e as discussed here
https://stackoverflow.com/questions/31642477/how-to-publish-an-npm-package-with-distribution-files

ryan-ludwig:
VUE-910

@kiva/kv-tokens/dist/scss/_variables.scss
ryan-ludwig: Let's put px in tokens file.

@kiva/kv-tokens/dist/css/variables.css
ryan-ludwig: Should add a github action to rebuild these anytime tokens file is changed.

PR #145, opened 2021-12-20
feat: kvcontentfulimg component was rewritten to use composition api

roger-in-kiva:


@kiva/kv-components/vue/KvContentfulImg.vue
```diff
@@ -172,23 +173,33 @@ export default {
...
+		const {
+			contentfulSrc,
+			width,
+			height,
+		} = toRefs(props);
+
+		const buildUrl = (image = null, multiplier = 1) => {
+			let src = image && image.url ? `${image.url}?` : `${contentfulSrc}?`;
+			const imgWidth = image ? image.width : width;
+			const imgHeight = image ? image.height : height;
```
emuvente, line 186: Because `contentfulSrc`, `width` and `height` are refs, their value needs to be accessed with `.value` when in the setup function, for example:
```
const imgWidth = image ? image.width : width.value;
```

roger-in-kiva: This was fixed, thank you!

emuvente (approved):
Nice!

PR #149, opened 2022-01-03
Vue3 kvcarousel vue 889

roger-in-kiva:
kvcarousel was rewritten to use composition api

This is a useful issue for this component
https://github.com/vuejs/composition-api/issues/586


ryan-ludwig:
Chromatic caught this diff which points to a behavior change. Previously clicking the arrow would scroll all the cards in view over. This version appears to only scroll one card.
Compare: https://608b4cf87f686c00213841b1-xdrnjqlbek.chromatic.com/?path=/story/kvcarousel--multiple-loan-cards-2 
to the old version:
https://608b4cf87f686c00213841b1-prfsdrulgh.chromatic.com/?path=/story/kvcarousel--multiple-loan-cards-2

@kiva/kv-components/tests/unit/specs/components/KvCarousel.spec.js
```diff
@@ -0,0 +1,61 @@
...
+
+	// 	await switchEl.click();
+	// 	expect(emitted().change[0]).toEqual([true]);
+
+	// 	await switchEl.click();
+	// 	expect(emitted().change[1]).toEqual([false]);
+	// 	expect(emitted().change.length).toBe(2);
+	// });
+
+	// it('has no automated accessibility violations', async () => {
```
emuvente, line 54: I think it would be good to have this test at a minimum, so that we can check for accessibility violations without much work.

@kiva/kv-components/vue/KvCarousel.vue
```diff
@@ -1,11 +1,12 @@
 <template>
 	<section
-		ref="KvCarousel"
+		ref="rootEl"
 		class="tw-overflow-hidden tw-w-full"
 		aria-label="carousel"
 	>
 		<!-- Carousel Content -->
 		<div
+			role="carousel"
```
emuvente, line 9: I don't think `carousel` is a recognized ARIA role, because I don't see it in this list https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles. I took a look at this W3C guide which doesn't appear to use roles at all for carousels https://www.w3.org/WAI/tutorials/carousels/structure/. 

Was this role added for the 'renders with a role of "carousel"' test? If so, I would recommend not implementing that test, as I don't think this carousel needs to have a role.

@kiva/kv-components/tests/unit/specs/components/KvCarousel.spec.js
roger-in-kiva: Is updated now.

@kiva/kv-components/vue/KvCarousel.vue
roger-in-kiva: Yeah, it was added for testing, but removed now.

emuvente (approved):
👍  Looks great!

ryan-ludwig (approved):

ryan-ludwig (approved):

emuvente:
@roger-in-kiva please work on getting the tests passing that I added to this PR. The tests can be run with `npm test -w test-vue2 -w test-vue3 KvCarousel`, and KvSwitch can be used as an example for getting the tests passing.

PR #150, opened 2022-01-04
feat: kvselect component was rewritten and basic test file was added

roger-in-kiva:
kvselect component was rewritten to use composition api and a basic test file was added


emuvente (commented):
I think this just needs the `emits: ['change', 'update:modelValue']` option defined, and then this should be good to go!

roger-in-kiva:
> I think this just needs the `emits: ['change', 'update:modelValue']` option defined, and then this should be good to go!

Fixed, thanks @emuvente !

@kiva/kv-components/vue/KvSelect.vue
```diff
@@ -6,6 +6,7 @@
 				:id="id"
 				:disabled="disabled"
 				:value="value"
+				role="select"
```
ryan-ludwig, line 9: This role can be removed as it is redundant on a `<select>` element

roger-in-kiva: Nice point, thanks. Is fixed.

@kiva/kv-components/vue/KvSelect.vue
```diff
@@ -58,21 +58,32 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		/**
```
ryan-ludwig, line 61: We didn't need this extra prop in the vue2 version. I think we can just use the value prop as before.

@kiva/kv-components/vue/KvSelect.vue
```diff
@@ -58,21 +58,32 @@ export default {
...
-	data() {
-		return {
-			mdiChevronDown,
-		};
-	},
-	methods: {
-		onChange(event) {
+	emits: [
+		'change',
+		'update:modelValue',
```
ryan-ludwig, line 71: this would be `'update:value'`

@kiva/kv-components/vue/KvSelect.vue
```diff
@@ -58,21 +58,32 @@ export default {
...
+		const onChange = (event) => {
 			/**
 			* The value that the select has changed to
 			* @event change
 			* @type {Event}
 			*/
-			this.$emit('change', event.target.value);
-		},
+			emit('change', event.target.value);
+			emit('update:modelValue', event.target.value);
```
ryan-ludwig, line 81: and `emit('update:value', event.target.value);`

@kiva/kv-components/vue/KvSelect.vue
ryan-ludwig: Ahh, thanks for the link. In that case I think we should remove the `value` prop

emuvente: Agreed, the `value` prop should be removed and `modelValue` should replace any usage of it. That will also mean that the model option should changed to:
```
model: {
	prop: 'modelValue',
	event: 'update:modelValue',
},
```

roger-in-kiva: Good!
The prop was removed and replaced with modelValue.

roger-in-kiva:
Ready for testing!

emuvente (approved):
Looks great!

PR #151, opened 2022-01-04
feat: kvtextlink component was rewritten to use composition api

roger-in-kiva:
kvtextlink component was rewritten to use composition api


ryan-ludwig (approved):

@kiva/kv-components/vue/KvTextLink.vue
```diff
@@ -73,44 +80,59 @@ export default {
...
-	},
-	methods: {
-		onClick(event) {
+		});
+
+		const onClick = (event) => {
 			// emit a vue event and prevent native event
 			// so we don't have to write @click.native in our templates
-			if (this.tag === 'button') {
+			if (tag === 'button') {
```
emuvente, line 107: I think this needs to check `tag.value` instead of `tag`

roger-in-kiva: Fixed, thanks!

emuvente (approved):
Looks good!

PR #153, opened 2022-01-05
misc Safari fixes

ryan-ludwig:
Will need to get these changes into the vue3 branch as well


mcstover (approved):
Does this address the sometimes giant icons in the loan card CTA?

ryan-ludwig:
> Does this address the sometimes giant icons in the loan card CTA?

@mcstover No, but that's likely the same issue of not having a width or height set on the `<kv-material-icon>`. Chrome and Firefox keep the svg icon at it's default size, but Safari makes it really big if you don't set `<kv-material-icon class="tw-w-3" />`

roger-in-kiva (approved):
Great!

emuvente (approved):
👍

PR #157, opened 2022-01-06
feat: kvswitch component was rewritten using composition api

roger-in-kiva:
kvswitch component was rewritten using composition api


emuvente (commented):
For this to work with v-model in Vue3, I think we need to rename the `checked` prop to be `modelValue`, and rename the `change` event to be `update:modelValue`. We would also need to change the model option so that Vue 2 picks up the changed names:
```
model: {
	prop: 'modelValue',
	event: 'update:modelValue',
},
```
Some of the stories may need to be updated for this component as well, so that they listen to the `update:modelValue` event instead of the `change` event.

roger-in-kiva:
> 

This means that the without V-Model option in storybook should be removed?

emuvente:
> This means that the without V-Model option in storybook should be removed?

I think we can keep that to demonstrate how it can be used with without v-model. The story template would have these lines removed:
```
:checked="switchExampleModel"
@change="(val) => switchExampleModel = val"
```
and replaced with:
```
:modelValue="switchExampleModel"
@update:modelValue="(val) => switchExampleModel = val"
```

roger-in-kiva:
> update:modelValue

Nice Suggestion, the feature was added.

emuvente:
This looks good to go! Thanks for making those updates!

I added some tests to check attribute and v-model behavior, and a fix for the attribute test failures so that it can be used for other components.

emuvente (approved):

mcstover (approved):

ryan-ludwig (approved):

PR #158, opened 2022-01-06
test: moves component tests to dedicated packages

emuvente:
This is the least complicated way I was able to get this working. By separating the testing for vue 2 and vue 3 into different packages, we're able to completely separate the dependencies and test configurations, making it very simple to setup each test environment. The most complex part of this is that the test packages have to manually copy the kv-components source files out of node_modules for babel & jest to properly recognize them.


ryan-ludwig (approved):

mcstover:
This is very clever. It seems now that there may be some incompatibility across the version of Testing Library. So I'm wondering if we can update the Vue 2 dep on v5.x to v6.x and it's new api.

roger-in-kiva:
Running well.
I tried to have an error modifying the button ref name, and it was not reported. It is just reporting a warning in kvtoast component.

roger-in-kiva (approved):

emuvente:
@mcstover I just added a util function for providing Vue Router to the Testing Library render function in the appropriate way for either Vue version, and it looks like that cleared up in the incompatibility issues.

PR #160, opened 2022-01-07
feat: kvtabpanel component was rewritten using compositon api and a b…

roger-in-kiva:
kvtabpanel component was rewritten using compositon api and a basic testing file was added


emuvente (approved):
Looks good!

PR #161, opened 2022-01-07
feat: kvtab component was rewritten using composition api

roger-in-kiva:
kvtab component was rewritten using composition api


emuvente (approved):
Looks good! Note: we should note this as a breaking change in the merge commit because of the switch from `for` to `forPanel` for the prop name.

PR #163, opened 2022-01-18
chore: use Vue 2 & composition api for storybook

emuvente:
Needed to add composition API to storybook config and ensure Vue 2 was being used for this to work properly.


roger-in-kiva (approved):
vue-demi tricks.
Great!

PR #164, opened 2022-01-19
handle $attrs differences between Vue 2 and 3 for KvSwitch VUE-935

emuvente:
This adds tests that reveal the different behavior of class, style, and event listener attributes between Vue 2 and 3, and adds a composable to help make the behavior consistent.


roger-in-kiva (approved):
Nice deconstruction of attrs.

mcstover:
Rad!

PR #167, opened 2022-01-24
VUE-933 fix: set correct module type for importing the package

ryan-ludwig:


mcstover:
Does this change anything with how UI or the other apps would import this lib?

ryan-ludwig:
I've done some testing today and the main drawback to the.cjs extension change is that anything that references those files will need to have their paths updated. In UI, that means `tailwind.config.cjs`, and `kivaColors.cjs`. For the Nuxt CMS, that'd be `tailwind.config.cjs`.

ryan-ludwig:
What I think may be a better solution is to convert all of the CommonJS in this repo to be ESModules. The Tailwind config starts the require chain, but I think it's doable and more future-proof.

ryan-ludwig:
I've attempted to convert tailwind.confg to EsModules, but have been unsuccessful. I'm not sure it's possible at this time.

ryan-ludwig:
If this goes in, we should set a breaking change flag since the import paths have changed.

ryan-ludwig:
Some extended reading: https://redfin.engineering/node-modules-at-war-why-commonjs-and-es-modules-cant-get-along-9617135eeca1

emuvente (approved):
I think this is a good way to make this work for now! That extended reading suggests a pretty complex setup that might be worth exploring in the future. We could also look into how other Vuejs libraries handle it.

As this should be a breaking change and we mainly need it for the Nuxt POC, I think we should merge this into the vue3 branch instead of main so that it can go into main at the same time as all of the breaking changes in the vue3 branch.

PR #172, opened 2022-02-03
chore!: tailwind v3 update

ryan-ludwig:
It seems I need some files changed within the package themselves for Lerna to know which packages need to be bumped


emuvente (approved):
:+1:

PR #176, opened 2022-02-15
fix(KvTokens): adjust prose styles to override new tailwind/typography defaults VUE-964

emuvente:


mcstover:
Pretty close! Just list elements that are a tad off now...

mcstover (approved):

PR #180, opened 2022-03-17
fix: return additional items for access as refs in vue2

mcstover:
I was able to see and access the `handleKeyDown` method from vue2 on the ref so taking the same approach with these.

I've been trying to find a way to test refs properly/successfully but haven't thus far.

### DIFF
--- a/@kiva/kv-components/tests/unit/specs/components/KvTabs.spec.js
+++ b/@kiva/kv-components/tests/unit/specs/components/KvTabs.spec.js
@@ -5,6 +5,8 @@ import '../../../../__mocks__/ResizeObserver';
 import KvTab from '../../../../vue/KvTab.vue';
 import KvTabPanel from '../../../../vue/KvTabPanel.vue';
 import KvTabs from '../../../../vue/KvTabs.vue';
+import KvButton from '../../../../vue/KvButton.vue';
+import addVueRouter from '../../utils/addVueRouter';
 
 // jsdom does not include scrollIntoView, so define it to avoid errors when switching tabs
 window.HTMLElement.prototype.scrollIntoView = () => {};
@@ -30,7 +32,6 @@ function renderOneTabSet() {
 		`,
 	});
 }
-
 describe('KvTabs', () => {
 	it('has no automated accessibility violations', async () => {
 		const { container } = renderOneTabSet();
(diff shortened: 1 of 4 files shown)


roger-in-kiva (approved):

emuvente (approved):
👍

mcstover:
Thanks for the advice @emuvente The added test and updated return of `tabContext` should be what we need to properly access the ref. At least the tests pass locally!

@kiva/kv-components/tests/unit/specs/components/KvTabs.spec.js
```diff
@@ -96,4 +97,74 @@ describe('KvTabs', () => {
...
+		expect(getAllByRole('tab').length).toBe(4);
+
+		// Click third tab and expect the third panel to be visible
+		await fireEvent.click(getByText('Third'));
+		expect(getByText('First Panel')).not.toBeVisible();
+		expect(getByText('Third Panel')).toBeVisible();
+		// Test access and updating of property ref
+		// expect the selected index value to be set to our new tab index
+		setTimeout(() => {
+			expect(selectedIndexCheck).toBe(2);
```
emuvente, line 157: Because this is in a setTimeout I think it may not be checked before the test completes. It probably wasn't working without this because of nextTick, right? A different way to test this would be to display the selected index in the test component and check for the existence of that using the async findBy* methods:
```
const { ..., findByText } = render({
	...
	data() {
		return { selectedIndex: null };
	},
	template: `
		<div data-testid="tabContainer">
			...
			<p>Tab index is {{ selectedIndex }}</p>
		</div>
	`,
	methods: {
		handleTabChanged() {
			this.$nextTick(() => {
				this.selectedIndex = this?.$refs?.tabsRef?.tabContext?.selectedIndex;
			});
		},
		...
});
...
await fireEvent.click(getByText('Third'));
expect(getByText('First Panel')).not.toBeVisible();
expect(getByText('Third Panel')).toBeVisible();
expect(findByText('Tab index is 2')).toBeVisible();
```

mcstover: Ok I'll test some other options tmrw. They were passing fine without the timeouts in vue 2 and the first one passed in vue3 but not the second, so I just wanted to add a delay.

mcstover: Ok, I was able to drop the next tick as well taking this route! 👍

mcstover:
Fixed up a couple consoles in the test logs that have been there a while
https://github.com/kiva/kv-ui-elements/runs/5221563417?check_suite_focus=true

PR #183, opened 2022-05-16
fix: label and input uuids didn't match

dyersituations:
Related to: https://kiva.atlassian.net/browse/VUE-1010

- Because uuids didn't match, the radio buttons weren't clickable unless the component was rendered late (timing issue)
- This is one of those one-line fixes that took a silly amount of time to track down

### DIFF
--- a/@kiva/kv-components/vue/KvRadio.vue
+++ b/@kiva/kv-components/vue/KvRadio.vue
@@ -151,7 +151,7 @@ export default {
 			modelValue,
 		} = toRefs(props);
 
-		let uuid = ref(`kvr-${nanoid(10)}`);
+		const uuid = ref(null);
 		const radioRef = ref(null);
 
 		const {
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/vue/KvRadio.vue
```diff
@@ -151,7 +151,7 @@ export default {
 			modelValue,
 		} = toRefs(props);
 
-		let uuid = ref(`kvr-${nanoid(10)}`);
+		const uuid = ref(null);
```
mcstover, line 154: i think we want to keep the setup init of the uuid as well. Making it a const should be fine though.

@kiva/kv-components/vue/KvRadio.vue
```diff
@@ -169,7 +169,7 @@ export default {
 		});
 
 		onMounted(() => {
-			uuid = `kvr-${nanoid(10)}`;
+			uuid.value = `kvr-${nanoid(10)}`;
```
mcstover, line 172: I think maybe the addition of `.value` is the culprit here.

dyersituations: Adding `.value` is indeed the fix.

roger-in-kiva (approved):

eddieferrer (approved):

mcstover (approved):

PR #185, opened 2022-05-27
refactor: use slideIndicator to hide carousel controls 

christian14b:
This pr aims to hide carousel controls when there is only 1 page to display.


mcstover (approved):
makes sense

emuvente (approved):
👍

PR #195, opened 2022-08-03
feat: add tailored emit event for MARS-199 experiment in kvLightbox

christian14b:
This pr modifies `lightbox-closed` event function for the purpose of tracking when a user clicks outside the modal to close it. With this, experiment MARS-199 should be able to track this click event.

### DIFF
--- a/@kiva/kv-components/vue/KvLightbox.vue
+++ b/@kiva/kv-components/vue/KvLightbox.vue
@@ -206,7 +206,6 @@ export default {
 	},
 	emits: [
 		'lightbox-closed',
-		'lightbox-closed-click-outside',
 	],
 	setup(props, { emit }) {
 		const {


@kiva/kv-components/vue/KvLightbox.vue
```diff
@@ -263,6 +264,8 @@ export default {
 
 		const onScreenClick = () => {
 			if (!preventClose.value) {
+				// This emit is only for MARS-199 experiment tracking and should be removed after results
```
dyersituations, line 267: Should this be temporary or just left in? It seems like a reasonable addition to the component. If so, this comment can be removed.

@kiva/kv-components/vue/KvLightbox.vue
```diff
@@ -206,6 +206,7 @@ export default {
 	},
 	emits: [
 		'lightbox-closed',
+		'lightbox-closed-clicking-outside',
```
dyersituations, line 209: Should this match `lightbox-closed-click-outside` below?

christian14b: Ops, you're right!

@kiva/kv-components/vue/KvLightbox.vue
```diff
@@ -263,6 +264,8 @@ export default {
 
 		const onScreenClick = () => {
 			if (!preventClose.value) {
+				// This emit is only for MARS-199 experiment tracking and should be removed after results
+				emit('lightbox-closed-click-outside');
```
eddieferrer, line 268: Is this really necessary? Can't we just track both close events with the same se_action param? Also, I'm pretty sure this will also fire the close event `emit('lightbox-closed');`

mcstover: If we wanted to keep just a single `lightbox-closed` event, which I do prefer, we could pass a payload along with the event and ensure each call to `hide(type)` includes the type.

then the emit would look something like `emit('lightbox-closed', { type: "button/close-x/background-click" })`

christian14b: I prefer that too! will work on that to make it cleaner

mcstover (approved):
Thanks!

dyersituations (approved):

emuvente (approved):

PR #197, opened 2022-08-10
Vue 1220 header create shared placeholder component

dyersituations:
https://kiva.atlassian.net/browse/VUE-1220

- Basic placeholder ported from `ui` repo

image

### DIFF
--- a/@kiva/kv-components/vue/KvLoadingPlaceholder.vue
+++ b/@kiva/kv-components/vue/KvLoadingPlaceholder.vue
@@ -1,0 +1,38 @@
+<template>
+	<div
+		class="
+			loading-placeholder
+			tw-w-full tw-h-full
+			tw-relative
+			tw-overflow-hidden
+			tw-bg-tertiary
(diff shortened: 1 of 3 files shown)


mcstover (approved):

mcstover:
@dyersituations I know I've approved already but should/could we add another story that helps cover the dynamic style, dynamic class and static class attributes that live on the original version? Seems like those attributes are needed now but just want to see if it's working ok...

dyersituations:
> @dyersituations I know I've approved already but should/could we add another story that helps cover the dynamic style, dynamic class and static class attributes that live on the original version? Seems like those attributes are needed now but just want to see if it's working ok...

When I ported this component over, I changed it from a `functional` component, since those aren't supported by Vue3. This is a simple enough component that I figured there shouldn't be a huge performance difference. I updated the story to show that both static and non-static versions of `class` and `style` work.

mcstover (approved):
Cool thanks!

PR #200, opened 2022-08-26
fix: ensure ref is defined in exported methods

dyersituations:
- Without this exposed ref, the same ref in the exposed methods appears to not be reactive
- Tested and verified locally in CPS using code in `node_modules`

### DIFF
--- a/@kiva/kv-components/vue/KvTextInput.vue
+++ b/@kiva/kv-components/vue/KvTextInput.vue
@@ -240,6 +240,7 @@ export default {
 			styles,
 			inputAttrs,
 			inputListeners,
+			textInputRef,
 		};
 	},
 };


mcstover (approved):
This should be ok but I'm a little concerned that this is used often in Ui. Maybe the Vue2 context was working??? Or maybe we weren't using refs much in concert with this component.

I found 3 usages that defined their own ref and 10 that had not, no usages of the `textInputRef` string.

dyersituations:
> This should be ok but I'm a little concerned that this is used often in Ui. Maybe the Vue2 context was working??? Or maybe we weren't using refs much in concert with this component.
> 
> I found 3 usages that defined their own ref and 10 that had not, no usages of the `textInputRef` string.

@mcstover I was wondering if it was an artifact of Vue3 importing this `vue-demi` component. At least one use case of this focus isn't working in the `ui` app -> if you open search in small view, it doesn't focus the textbox, even though there is code that is _supposed_ to do that. An alternative is to not worry about the auto-focusing when opening the search in small view and closing this PR. Thoughts?

mcstover:
I think we can go forward with it if it's behaving properly in your context with this added. From what I've seen it may fix what you've found in Ui or otherwise wasn't in use.

PR #202, opened 2022-09-07
feat: add vertical tab button for new home page exp

christian14b:
This pr adds a vertical tab button for [Home Page Experiment component](https://github.com/kiva/ui/pull/4162)

image

### DIFF
--- a/@kiva/kv-components/vue/KvButton.vue
+++ b/@kiva/kv-components/vue/KvButton.vue
@@ -175,14 +175,6 @@ export default {
 						classes = `${classes} tw-bg-caution hover:tw-bg-caution-highlight`;
 					}
 					break;
-				case 'vertical-tab':
-					classes = 'tw-text-h2 tw-border-transparent tw-bg-transparent tw-border-transparent';
-					if (state.value === 'active') {
-						classes = `${classes} md:tw-border-l-2 md:tw-border-l-action-highlight tw-text-action-highlight tw-rounded-none`;
-					} else {
-						classes = `${classes} hover:tw-bg-primary-inverse tw-text-primary-inverse`;
-					}
-					break;
 			}
 			return classes;
 		});
--- a/@kiva/kv-components/vue/stories/KvButton.stories.js
+++ b/@kiva/kv-components/vue/stories/KvButton.stories.js
@@ -7,7 +7,7 @@ export default {
 		variant: {
 			control: {
 				type: 'select',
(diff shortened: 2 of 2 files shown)


dyersituations (approved):

eddieferrer:
@mcstover @dyersituations 

I think that semantically, this is not a button and this is overloading the button component with too many variations. 

I would much rather see this implemented as a vertical variation of the KvTabs component. That supports a set of 1 tab (looks attached picture) or even multiple vertical tabs.

dyersituations:
> @mcstover @dyersituations
> 
> I think that semantically, this is not a button and this is overloading the button component with too many variations.
> 
> I would much rather see this implemented as a vertical variation of the KvTabs component. That supports a set of 1 tab (looks attached picture) or even multiple vertical tabs.

Good call, and I guess it would depend on the full use case. That's part of why I'm always hesitant to approve PRs without the full context. The code here seems fine, but without seeing the design/AC, it's hard to know if this is a button vs a tab vs something else entirely.

mcstover:
@eddieferrer I think that's a valid comment. @christian14b What do you think the lift would be to make alterations to the tabs component?

eddieferrer:
Looks great, don't forget to clean up KvButton and story

eddieferrer (approved):

@kiva/kv-components/vue/stories/KvTabs.stories.js
```diff
@@ -5,6 +5,9 @@ import KvTabPanel from '../KvTabPanel.vue';
 export default {
 	title: 'KvTabs',
 	component: KvTabs,
+	args: {
+		vertical: true,
```
dyersituations, line 9: I'm guessing you don't want vertical for all of the stories. If so, add to the `VerticalOrientation` story below.

dyersituations (approved):

@kiva/kv-components/vue/stories/KvTabs.stories.js
mcstover: Agree, please verify that behavior and appearance of the existing stories.

mcstover: Well, it looks like our other existing stories aren't "applying" the default setting to their props. Thus, there were no changes to these stories in the Chromatic diff.

mcstover (approved):

PR #203, opened 2022-09-09
fix(KvCarousel): incorrect slide marked as current

emuvente:
Noticed this while working on the story card carousel. This bug causes the next slide to be marked as current, instead of the current slide.

### DIFF
--- a/@kiva/kv-components/vue/KvCarousel.vue
+++ b/@kiva/kv-components/vue/KvCarousel.vue
@@ -1,7 +1,7 @@
 <template>
 	<section
 		ref="rootEl"
-		class="tw-overflow-hidden tw-w-full"
+		class="kv-carousel tw-overflow-hidden tw-w-full"
 		aria-label="carousel"
 	>
 		<!-- Carousel Content -->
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/vue/KvCarousel.vue
```diff
@@ -29,7 +29,7 @@
 		<!-- Carousel Controls -->
 		<div
 			v-if="slideIndicatorCount > 1"
-			class="tw-flex
+			class="kv-carousel__controls tw-flex
```
dyersituations, line 32: Are these two added classes defined somewhere else?

dyersituations (approved):

mcstover (approved):

@kiva/kv-components/vue/KvCarousel.vue
emuvente: I added these so I could reposition the controls in this pr https://github.com/kiva/ui/pull/4169

edwinf-barreram (approved):

PR #204, opened 2022-09-15
fix(KvCarousel): slidesToScroll should not be 0 MARS-248

emuvente:
Resizing the window when there is an invisible KvCarousel in the page currently causes the javascript to hang and the page to crash. The cause of the issue appears to be that slidesToScroll is set to 0 (the number of visible slides) when the carousel is reinitialized. Setting it as 'auto' instead in those cases made the issue go away in testing.


dyersituations (approved):

mcstover (approved):
Nice find!

PR #206, opened 2022-09-19
chore(deps): security updates for development module dependencies

emuvente:
Result of running `npm update url-parse minimist follow-redirects moment terser parse-url @actions/core`, based on the open dependabot pull requests.


mcstover (approved):
Thank you!

PR #211, opened 2022-09-27
fix(KvContentfulImg): match size and fit of picture to figure wrapper MARS-265

emuvente:
Are there landing pages we can look at to verify these changes work correctly?


eddieferrer (approved):
Looks good. I think marketing may have a better idea of active landing pages

edwinf-barreram (approved):

mcstover:
There seem to be some issues with the bg image on the  current homepage in dev... this could fix that up.

PR #212, opened 2022-09-29
feat: created and implemented a reusable global component and story (…

tiarascoleman:
…KvExpandableQuestion)[VUE-1318]

[VUE-1318]: https://kiva.atlassian.net/browse/VUE-1318?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/@kiva/kv-components/vue/KvExpandableQuestion.vue
+++ b/@kiva/kv-components/vue/KvExpandableQuestion.vue
@@ -81,13 +81,12 @@ export default {
 		const {
 			title,
 			active,
-			faq,
 		} = toRefs(props);
 		const open = ref(active.value || false);
 		const titleSlugified = computed(() => paramCase(title.value));
 
 		const toggleFAQ = () => {
-			props.kvTrackFunction(faq, 'toggle', titleSlugified, open.value ? 'expand' : 'collapse');
+			props.kvTrackFunction('faq', 'toggle', titleSlugified, open.value ? 'expand' : 'collapse');
 			open.value = !open.value;
 			emit('toggle', { title: titleSlugified });
 		};
--- a/@kiva/kv-components/vue/KvExpandableQuestion.vue
+++ b/@kiva/kv-components/vue/KvExpandableQuestion.vue
@@ -1,0 +1,124 @@
+<template>
+	<div>
+		<button
+			class="tw-w-full tw-py-2 tw-flex tw-justify-between"
+			@click="toggleFAQ"
+		>
+			<h3 class="tw-text-subhead tw-text-left">
+				{{ title }}
+			</h3>
+			<kv-material-icon
+				class="tw-w-4 tw-h-4"
+				:icon="open ? mdiChevronUp : mdiChevronDown"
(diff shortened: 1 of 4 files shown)


@kiva/kv-components/vue/KvExpandableQuestion.vue
```diff
@@ -0,0 +1,124 @@
...
+		 * Question Content - can accept raw html
+		* */
+		content: {
+			type: String,
+			default: '',
+		},
+		/**
+		 * Analytics Category
+		* */
+		analyticsCategory: {
```
eddieferrer, line 67: We can remove this prop. The analytics category for this should always be `faq`

mcstover (approved):
Just wanted to note that we've removed the "dependency" on access to the route hash and kvTrackEvent methods by making them props that are passed in when needed.

@kiva/kv-components/vue/KvExpandableQuestion.vue
mcstover: Nice!. @tiarascoleman note that is all lowercase `faq` which can be placed directly in the code.

mcstover: @eddieferrer any thoughts or changes we should make to the title in the analytics event?

@kiva/kv-components/vue/KvExpandableQuestion.vue
```diff
@@ -0,0 +1,124 @@
...
+			title,
+			active,
+			analyticsCategory,
+		} = toRefs(props);
+		const open = ref(active.value || false);
+		const titleSlugified = computed(() => paramCase(title.value));
+
+		const toggleFAQ = () => {
+			if (!open.value) {
+				props.kvTrackFunction(analyticsCategory.value, 'click-faq-expand', title.value);
```
eddieferrer, line 98: The category should be `faq`
The action should be `toggle`
The label can be the titleSlugified
The property can be used to indicate the state of the question maybe `expand` or `shrink`

So this can be `kvTrackFunction(`faq`, 'toggle', titleSlugified, <expand or shrink>`

@kiva/kv-components/vue/KvExpandableQuestion.vue
eddieferrer: See my other comment

@kiva/kv-components/vue/KvExpandableQuestion.vue
mcstover: Cool I think that makes sense. 

@tiarascoleman you should be able to assign that last property conditionally based on the value of `open.value` with a ternary operator.

Also if we're going to track both the expand and shrink states, we should remove the `if (!open.value) {}` block around the tracking call.

mcstover (approved):

dyersituations (approved):

PR #214, opened 2022-10-14
fix: turn off pre-wrap whitespace for KvExpandableQuestion and prose …

mcstover:
…for the title of the component

When these are embedded in a Rich Text Editor a wrapping parent div includes tw-prose and tw-whitespace-pre-wrap which creates a lot of extra unwanted padding around the elements.

The updates prevent that from applying where it shouldn't be.

Before:
image

After:
image


tiarascoleman (commented):
Looks good!

tiarascoleman (approved):

PR #223, opened 2022-12-01
fix: use woff2 font files

dyersituations:
- Changed typography to use the same fonts as the `ui`, `auth0`, and `cms-page-server` repos (woff2)
- It looks like typography used to have `MediumItalic`, but I'm only seeing that file loaded directly from the project folder in https://github.com/kiva/ui/blob/main/server/util/live-loan/live-loan-draw.js

### DIFF
--- a/@kiva/kv-tokens/configs/kivaTypography.cjs
+++ b/@kiva/kv-tokens/configs/kivaTypography.cjs
@@ -18,36 +18,36 @@ const webFonts = [
 		'@font-face': {
 			fontFamily: 'PostGrotesk',
 			fontWeight: '400',
+			fontStyle: 'italic',
+			fontDisplay: 'swap',
+			src: 'url(//www-kiva-org.freetls.fastly.net/static/fonts/PostGrotesk-BoldItalic.ace8261.woff2) format(\'woff2\')',
+		},
+	},
+	{
+		'@font-face': {
+			fontFamily: 'PostGrotesk',
+			fontWeight: '300',
 			fontStyle: 'normal',
 			fontDisplay: 'swap',
-			src: 'url(//www-kiva-org.freetls.fastly.net/static/fonts/PostGrotesk-Medium.592afe4.woff) format(\'woff\')',
+			src: 'url(//www-kiva-org.freetls.fastly.net/static/fonts/PostGrotesk-Book.246fc8e.woff2) format(\'woff2\')',
+		},
+	},
+	{
+		'@font-face': {
+			fontFamily: 'PostGrotesk',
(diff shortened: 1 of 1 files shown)


mcstover:
Yea the prose demo is def looking crazy, especially at the bottom lolz. However, that's expected with the change from using medium italic to bold italic for the font file. It looks fine and better in some cases so as long as that was intended we should be fine.

dyersituations:
> Yea the prose demo is def looking crazy, especially at the bottom lolz. However, that's expected with the change from using medium italic to bold italic for the font file. It looks fine and better in some cases so as long as that was intended we should be fine.

My assumption is that the change is expected, since we don't use medium italic anywhere but `live-loan-draw`, but I could be missing something important!

emuvente:
Was there a request from Creative to make the italics bolder? I don't think we should be doing that for no reason.

dyersituations:
> Was there a request from Creative to make the italics bolder? I don't think we should be doing that for no reason.

We just use the bold italics on the other repos, and I don't know the history of why. Why don't I just revert the italics change, because if that file is requested by the app CSS, I don't think it'll be loaded at all. And the apps already load the bold italic.

emuvente:
I did some investigating. It turns out there was [a typo introduced back in July](https://github.com/kiva/ui/commit/2cb92c921df015d0af552977fe623f4db5a976c4#diff-cbdeb1437d0d4387bb490591d303f56a5e8d58fdf06dcd692b2d31e165a6d6ddR30) that changed the name of the 400 italics font to be `BoldItalic` instead of `MediumItalic`. The correct one to be using is `MediumItalic`. We should fix the apps to use it instead.

dyersituations:
> I did some investigating. It turns out there was [a typo introduced back in July](https://github.com/kiva/ui/commit/2cb92c921df015d0af552977fe623f4db5a976c4#diff-cbdeb1437d0d4387bb490591d303f56a5e8d58fdf06dcd692b2d31e165a6d6ddR30) that changed the name of the 400 italics font to be `BoldItalic` instead of `MediumItalic`. The correct one to be using is `MediumItalic`. We should fix the apps to use it instead.
Interesting, thanks for the exploration. I can update the apps after the package gets released. Working through some other work ATM, but I'll get this done in the next couple business days!

dyersituations:
@mcstover @emuvente Looks like we'll need to deploy `ui` with `MediumItalic` being used before we get the hash for the file. So this PR will stay open until after the next `ui` prod release.

mcstover (approved):

emuvente (approved):

PR #228, opened 2023-01-20
Core 933 kvtoast update

roger-in-kiva:
slot element added to kvToast component to inject code from parent to tracking and emitting events mainly


@kiva/kv-components/vue/KvToast.vue
```diff
@@ -71,6 +71,10 @@
 							tw-flex
 						"
 						>
+							<slot
+								v-if="hasParentContentSlot"
+								name="parentContent"
```
dyersituations, line 76: Calling this slot something like `toastContent` might make more sense.

@kiva/kv-components/vue/KvToast.vue
```diff
@@ -71,6 +71,10 @@
...
+							></slot>
 							<!-- eslint-disable vue/no-v-html -->
 							<p
```
dyersituations, line 79: The slot won't replace this message content entirely? Like either the component displays the slot or the old simple message string.

roger-in-kiva: There are other usages with the message prop, in that case we need to propagate this update to them or what do you suggest?

@kiva/kv-components/vue/KvToast.vue
roger-in-kiva: done

@kiva/kv-components/vue/KvToast.vue
dyersituations: I just meant that do we have a usage that needs _both_ the slot and the message prop? I was wondering if we needed a `v-else` on the `p` tag here.

roger-in-kiva: good call! as the slot is new, we gonna suppose that there is not a usage that needs both, ill add the `v-else` to p element as suggested, ty

dyersituations (approved):

roger-in-kiva:
@emuvente @michelleinez any thought about this?

PR #231, opened 2023-02-10
feat: hide delay param for toast

dyersituations:
https://kiva.atlassian.net/browse/CORE-1052

- We wanted a way to show a toast for a specific amount of time
- This allows us to tell a toast to display for exact 10s, for example


mcstover (approved):
Cool

PR #235, opened 2023-04-14
feat: move KvPagination component to kv-elements

christian14b:
Moving KvPagination component for blog work.

### DIFF
--- a/@kiva/kv-components/tests/unit/specs/components/KvPagination.spec.js
+++ b/@kiva/kv-components/tests/unit/specs/components/KvPagination.spec.js
@@ -1,0 +1,142 @@
+import { render } from '@testing-library/vue';
+import userEvent from '@testing-library/user-event';
+import KvPagination from '../../../../vue/KvPagination.vue';
+
+global.scrollTo = jest.fn();
+
+describe('KvPagination', () => {
+	afterEach(jest.clearAllMocks);
+
+	it('should render arrows disabled by default', async () => {
+		const user = userEvent.default.setup();
+
+		const { getByLabelText, emitted } = render(KvPagination, { props: { limit: 10, total: 0 } });
+
+		await user.click(getByLabelText('Previous page'));
+		await user.click(getByLabelText('Next page'));
+
+		expect(emitted()['page-changed']).toBe(undefined);
+	});
+
+	it('should render aria current page label', () => {
+		const { getByText } = render(KvPagination, { props: { limit: 10, total: 30 } });
+
+		getByText("You're on page");
+	});
+
+	it('should render fewer pages', () => {
+		const { getByLabelText } = render(KvPagination, { props: { limit: 10, total: 30 } });
+
+		getByLabelText('Page 1');
+		getByLabelText('Page 2');
+		getByLabelText('Page 3');
+	});
+
+	it('should render more pages', () => {
+		const { getByLabelText } = render(KvPagination, { props: { limit: 10, total: 1000 } });
+
+		getByLabelText('Page 1');
+		getByLabelText('Page 2');
+		getByLabelText('Page 3');
+		getByLabelText('Page 4');
+		getByLabelText('Page 100');
+	});
+
+	it('should render fourth selected', () => {
+		const { getByLabelText } = render(
+			KvPagination, { props: { limit: 10, total: 1000, offset: 30 } },
+		);
+
+		getByLabelText('Page 1');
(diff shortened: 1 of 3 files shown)


@kiva/kv-components/vue/KvPagination.vue
```diff
@@ -0,0 +1,198 @@
...
+			if (this.scrollToTop && window.scrollTo) {
+				window.scrollTo({ top: 0, left: 0, behavior: 'smooth' });
+			}
+
+			this.$emit('page-changed', { pageOffset: number * this.limit });
+		},
+		clickPage(number) {
+			this.pageChange(number);
+
+			this.kvTrackFunction(this.trackEventCategory, 'click-page-pager', number + 1);
```
eddieferrer, line 180: Would this be a good time to transition these events to our latest web event guidelines? The actions in particular should be 'click' -- So for this event I would do: 
`this.trackEventCategory, 'click', 'page-pager', number + 1`

dyersituations: I guess it depends on if we have reports that depend on these. My guess is we don't, but I don't know for sure...

dyersituations (approved):

dyersituations:
Approved pending test failures...

@kiva/kv-components/vue/KvPagination.vue
mcstover: Thank you @eddieferrer That is a good point and I agree we should split the events in the way you have described. `click` or other appropriate action is the 2nd param, followed by a `label` in the 3rd position and then a `value` where appropriate.

mcstover: I think we should use "pagination-number" instead of "page-pager" in this case.

@kiva/kv-components/vue/KvPagination.vue
```diff
@@ -0,0 +1,198 @@
...
+			this.pageChange(previous);
+
+			this.kvTrackFunction(this.trackEventCategory, 'click-previous-pager', previous + 1);
+		},
+		clickNext() {
+			const next = this.current + 1;
+
+			this.pageChange(next);
+
+			this.kvTrackFunction(this.trackEventCategory, 'click-next-pager', next + 1);
```
mcstover, line 194: For previous and next please use "click" for the action followed by "pagination-previous" or "pagination-next" and the value after that.

@kiva/kv-components/vue/KvPagination.vue
mcstover: @eddieferrer Do you think we should go ahead and duplicate the "number" portion of this as the 5th param?

eddieferrer: @mcstover I agree with the `pagination-` as the label. For this event maybe: 
`this.trackEventCategory, 'click', 'pagination-number', null, number + 1` would work. But I'm also ok with: 
`this.trackEventCategory, 'click', 'pagination-number', number + 1` just keeping in mind that the 4th param should be a string.

@kiva/kv-components/tests/unit/specs/components/KvPagination.spec.js
```diff
@@ -0,0 +1,132 @@
...
+
+global.scrollTo = jest.fn();
+
+describe('KvPagination', () => {
+	afterEach(jest.clearAllMocks);
+
+	it('should render arrows disabled by default', async () => {
+		const { getByLabelText, emitted } = render(KvPagination, { props: { limit: 10, total: 0 } });
+
+		await userEvent.click(getByLabelText('Previous page'));
```
dyersituations, line 13: Last I heard `user.setup` was the "new" approach to click events in `testing-library`. Has the recommendation changed?

christian14b: Yeah, that's right. The problem is that the project is using `13.2.1` . I tried upgrading it to `14.2.0` version, but the upgrade caused some weird errors in KvTextInput and KvTabs tests

mcstover: I we will have to upgrade the lib which comes with some refactoring of other tests to move to the new approach.

dyersituations: Sounds good, thanks.

mcstover (approved):

PR #236, opened 2023-04-26
Core 1201 refactor qf to use loan search utilities

dyersituations:
https://kiva.atlassian.net/browse/CORE-1309

**Overview**

- New package `@kiva/kv-loan-filters` contains the filtering modules used to previously power the lend/filter page
- The new modules can be used to power all filtering UI moving forward
- This package was created while simulataneously implementing changes in the `ui` repo to use it - any new changes were due to abstractions to support filters in both /lend/filter and /lending-home quick filters

**Package changes**

- Moved linting to packages from root to allow different linting per package (in this case needed TS linting)
- Added build stage needed for TS packages
- Resolved a few minor linting errors in components package
- Updated lerna to latest version

**@kiva/kv-loan-filters specific notes**

- TypeScript output generated via `npm run build`
- Output needed both ES module and CommonJS support - ES module for the Vue 2 + 3 frontends, CommonJS for the `ui` node server - `tsup` is a 3rd party TypeScript compiler library that supports creating these two outputs
- The only 3rd party dependency in the output is currently `lodash.orderby`, but there's a ticket below to consider replacing it
- All of the 19 filter modules were ported from `ui` repo (includes one new module for a categories filter)
- Filter modules changed slightly based on further abstraction needed
- Ported utilities specific to loan filters w/ their existing unit tests
- Moved some component-specific utilities to the shared utils to make importing back into `ui` cleaner
- Package exports utils, util configs (used for UI interpretation), and filter utils (general filter modules powering functionality)
- Ported channel map file, since it's used for loan filtering

**Future work**

https://kiva.atlassian.net/browse/CORE-1311

- Adding typing throughout the new package - held off due to the number of changes in this initial port, all files are .ts, but there are no types explicitly added yet
- Potentially porting graphql files for loan search state (requires adding linting for .graphql)
- Consider replacing lodash.orderby with custom solution
- Port Apollo watchQuery to loan filters package
- Investigate allowing sub directory import from package - this was causing havoc initially with compilation
- Review/add unit tests - noticed some tests missing during port


roger-in-kiva (approved):
Nice work! TY.

@kiva/kv-tokens/package.json
```diff
@@ -7,7 +7,9 @@
     },
     "main": "index.js",
     "scripts": {
-        "test": "echo No tests specified for @kiva/kv-tokens."
+        "test": "echo No tests specified for @kiva/kv-tokens.",
+		"lint": "eslint --ext .js ./",
+		"build": "echo No build needed for @kiva/kv-tokens."
```
mcstover, line 12: spacing

mcstover (approved):
Very nice!

mcstover:
There may be some opportunity to move a couple of the utils here into a more general space at some point but this is s great first utility library!

@kiva/kv-tokens/package.json
dyersituations: Oops, missed one, thanks. Fixed!

dyersituations:
> There may be some opportunity to move a couple of the utils here into a more general space at some point but this is s great first utility library!

Yeah definitely. This is meant to be a first pass at a fairly large group of files. We can arrange, create other packages, etc. I plan on getting another package relatively soon for the experimentation stuff.

emuvente (approved):
This is great! I like the use of tsup!

PR #237, opened 2023-05-03
feat: add custom blog colors

christian14b:
Custom colors for blog design

### DIFF
--- a/@kiva/kv-tokens/configs/tailwind.config.cjs
+++ b/@kiva/kv-tokens/configs/tailwind.config.cjs
@@ -46,6 +46,8 @@ module.exports = {
 			marigold: colors.marigold,
 			'desert-rose': colors['desert-rose'],
 			stone: colors.stone,
+			gray: colors.gray,
+			social: colors.social,
 			// themable colors are defined in the 'extend' section as custom properties
 		},
 		spacing: {
--- a/@kiva/kv-tokens/configs/tailwind.config.cjs
+++ b/@kiva/kv-tokens/configs/tailwind.config.cjs
@@ -161,7 +161,10 @@ module.exports = {
 			placeholderColor: buildColorChoices('text'),
 			backgroundColor: buildColorChoices('background'),
 			gradientColorStops: buildColorChoices('background'),
-			borderColor: buildColorChoices('border'),
+			borderColor: {
+				DEFAULT: 'currentColor',
+				...buildColorChoices('border'),
+			},
 			divideColor: buildColorChoices('border'),
 			ringColor: buildColorChoices('border'),
 			ringOffsetColor: buildColorChoices('border'),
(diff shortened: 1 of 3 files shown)


@kiva/kv-tokens/configs/tailwind.config.cjs
```diff
@@ -46,6 +46,8 @@ module.exports = {
 			marigold: colors.marigold,
 			'desert-rose': colors['desert-rose'],
 			stone: colors.stone,
+			gray: colors.gray,
```
mcstover, line 49: I think this will open up several "grays" that were previously relegated to various other tw classes such as border.

mcstover (approved):

@kiva/kv-tokens/configs/tailwind.config.cjs
dyersituations: Exciting!

dyersituations (approved):

emuvente (approved):

mcstover:
@christian14b This appears to have affected and changed the styles for the default `tw-border` class. I noticed this by reviewing the chromatic diff for the theme switcher. Can you please investigate before we merge this?

mcstover (changes requested):
This somehow breaks the styles for `tw-border` class so we need to ensure that remains the same as it's widely used in it's current form.

mcstover (approved):
With a little more investigation it seems that something has changed with how the "default/unspecified" border color is assigned, _possibly in the way storybook handles setting variables_. The published version of storybook shows that the border color is assigned to `currentColor` whereas the new version shows it using #E0E0E0 which is not defined as a border color. Rather that value is from tertiary background or gray 200. _So maybe gray 200 is set as a default but I couldn't find that._

In most instances we actually set a border color explicity when applying our classes, there were only 2 places in UI without a color. The changes exhibited here show what happens when a border color is not specified. The border set in the Theme Provider stories just uses `tw-border` with no color specified which falls back to this arbitrary default.

I think this is fine and won't introduce much disruption.

mcstover:
With this update we may see a change to:
- KvComponent: The border color underneath he KvAccordian items
- UI: border color in `ui/src/pages/Settings/RecoveryCodeConfirm.vue`

christian14b:
@mcstover. I was digging into that too [repo](https://github.com/tailwindlabs/tailwindcss/pull/6926/files#diff-d0363b2e64f7e7cdb4b9ba38cbc5ce90e0aa48eedc8ebd33afe83bb7627d856cR12). Apparently border-color is being override because of tailwind default (Not sure about this).

Either way, I tried adding a DEFAULT value to borderColor in `tailwind.config.cjs` and seems to solve to problem. I'll leave it here as a possible solution:

```
borderColor: {
	DEFAULT: 'currentColor',
	...buildColorChoices('border'),
},
```

mcstover:
@christian14b I think adding that default is a good idea and would keep things as they were.

mcstover (approved):

PR #242, opened 2023-05-16
feat: remove countdown for now from loan tags

dyersituations:
- Turns out the vue countdown component we're using encounters an error in Vue 3 (we have 600+ errors in Sentry since release yesterday)
- Turning off countdown (checked with product) and creating a ticket for sometime in the next few weeks to fix
- Loan tag still appears but just says "Ending soon"

### DIFF
--- a/@kiva/kv-components/vue/KvLoanTag.vue
+++ b/@kiva/kv-components/vue/KvLoanTag.vue
@@ -5,7 +5,8 @@
 		style="color: #CE4A00;"
 	>
 		{{ tagText }}
-		<vue-countdown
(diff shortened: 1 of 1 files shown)


eddieferrer (approved):
Looks like the version was just copied from UI repo. You'll need version ^2 to get vue3 support

dyersituations:
> Looks like the version was just copied from UI repo. You'll need version ^2 to get vue3 support

Yeah I did see that in the 3rd party repo, but we'll need to make sure it's also backwards compatible... I created a ticket for us to investigate soon, but I want to do a quick hotfix.

PR #243, opened 2023-05-18
feat: changes needed for implicit typing in CPS

dyersituations:
- Some temp typing changes (and one good permanent change) for implicit typing in CPS when using these methods
- Soon I'll add typing to the package, but for now these `any` entries play nicely with a TS project importing the package


roger-in-kiva (approved):
Looks useful for now!
Should you create a tech ticket for future work?

dyersituations:
> Looks useful for now! Should you create a tech ticket for future work?

Yup there's already a ticket.

PR #245, opened 2023-05-23
Core 1372 filters location selector

roger-in-kiva:
- new `KvLocationSelector` component added to `kv-components` package to have a shared countries checkbox list

### DIFF
--- a/@kiva/kv-components/vue/KvLocationSelector.vue
+++ b/@kiva/kv-components/vue/KvLocationSelector.vue
@@ -1,30 +1,38 @@
 <template>
 	<form
 		class="tw-grid tw-gap-1.5 tw-mb-2"
-		:class="`tw-grid-cols-${numberColumns}`"
 		@submit.prevent
 	>
+		<div class="tw-text-right tw-hidden md:tw-block tw-mb-1 tw-pr-2">
+			Locations
+		</div>
 		<fieldset
 			v-for="option in displayedRegions"
 			:key="option.region"
+			class="md:tw-border-b tw-border-tertiary md:tw-pb-1 md:tw-mb-1"
 		>
-			<legend>
+			<legend class="tw-w-full md:tw-mb-1">
 				<button
-					class="tw-transition-all"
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/vue/KvLocationSelector.vue
```diff
@@ -0,0 +1,236 @@
...
+		 *     isoCode: 'US',
+		 *     numLoansFundraising: 1,
+		 *   }]
+		 * }]
+		 */
+		regions: {
+			type: Array,
+			default: undefined,
+		},
+		activeIsoCodes: {
```
dyersituations, line 119: This can be a follow-up task, please add a ticket, but we'll need this component to watch `activeIsoCodes` for changing the selected values based on this prop changing. The use case:

1. User selects a country not in US
2. User selects US category
3. Filters get updated to have country selection that matches category
4. Location filter should show the updated filter (without user having changed the location filter)

There are other cases, but that's the general idea.

dyersituations (approved):

mcstover:
Is there a way to clear up some of those lint warnings in the PR... Seems a bit excessive.

dyersituations:
> Is there a way to clear up some of those lint warnings in the PR... Seems a bit excessive.

Oh yeah, those are my fault. I have a ticket to address the typing in the loan filters package, which should resolve all of the current linting errors in the filter modules.

mcstover (approved):

@kiva/kv-components/vue/KvLocationSelector.vue
```diff
@@ -213,6 +214,11 @@ export default {
 			return total;
 		};
 
+		watch(activeIsoCodes, (val) => {
+			selected.value = val;
```
dyersituations, line 218: We should probably check that the arrays of values don't have the exact same values before applying.

roger-in-kiva: Cool! Even, I'll add a watcher for the regions prop that could be changed by other filters.

dyersituations (approved):

PR #246, opened 2023-05-24
feat: add new package kv-shop for payment and checkout logic

emuvente:
A lot of TODOs and several open questions in this, but I wanted to start to get eyes on it.

### DIFF
--- a/@kiva/kv-shop/.eslintrc.cjs
+++ b/@kiva/kv-shop/.eslintrc.cjs
@@ -1,0 +1,29 @@
+module.exports = {
+	parser: 'vue-eslint-parser',
+	parserOptions: {
+		parser: '@typescript-eslint/parser',
+	},
+	plugins: [
+		'@typescript-eslint',
+		'vue',
+	],
+	extends: [
+		'../../.eslintrc.cjs',
+		'plugin:@typescript-eslint/eslint-recommended',
+		'plugin:@typescript-eslint/recommended',
+		'plugin:import/typescript',
+		'plugin:vue/recommended',
+		'@vue/typescript/recommended',
+	],
+	rules: {
+		'vue/html-indent': ['error', 'tab'],
+		// disallow non-void html elements to be self-closing
+		'vue/html-self-closing': ['error', {
+			html: {
+				normal: 'never',
+			},
+		}],
+		// allow imports without file extensions
+		'import/extensions': 'off',
+	},
+};
--- a/@kiva/kv-shop/package.json
+++ b/@kiva/kv-shop/package.json
@@ -1,0 +1,53 @@
+{
+    "name": "@kiva/kv-shop",
+    "version": "1.0.0",
+    "type": "module",
+    "publishConfig": {
+        "access": "public"
+    },
+    "main": "dist/index.cjs",
+    "module": "dist/index.js",
+    "types": "dist/index.d.ts",
+    "files": [
+        "dist"
+    ],
+    "exports": {
+        ".": {
+            "require": "./dist/index.cjs",
+            "import": "./dist/index.js",
+            "types": "./dist/index.d.ts"
+        }
+    },
+    "scripts": {
+        "build": "tsup src/**/*.ts --format cjs,esm --dts --clean && cp -R src/components/ dist/components/",
+        "lint": "eslint --ext .ts,.vue ./src",
+        "test": "echo No test yet for kv-shop"
+    },
+    "devDependencies": {
+        "@typescript-eslint/eslint-plugin": "^5.59.7",
+        "@typescript-eslint/parser": "^5.59.7",
+        "@vue/eslint-config-typescript": "^11.0.3",
+        "eslint-plugin-vue": "^9.14.0",
+        "tsup": "^6.7.0",
+        "typescript": "^5.0.4",
+        "vue": "2.6",
+        "vue-eslint-parser": "^9.3.0"
+    },
+    "dependencies": {
+        "@apollo/client": "^3.7.14",
+        "@kiva/kv-components": "^3.19.0",
+        "@types/braintree-web-drop-in": "^1.34.2",
+        "braintree-web-drop-in": "^1.37.0",
+        "numeral": "^2.0.6",
+        "vue-demi": "^0.14.1"
+    },
+    "peerDependencies": {
+        "@vue/composition-api": "^1.0.0-rc.1",
+        "vue": "^2.0.0 || >=3.0.0"
+    },
+    "peerDependenciesMeta": {
+        "@vue/composition-api": {
+            "optional": true
+        }
+    }
+}
--- a/@kiva/kv-shop/src/components/KvPaymentSelect.vue
+++ b/@kiva/kv-shop/src/components/KvPaymentSelect.vue
@@ -1,0 +1,351 @@
+<template>
+	<div class="kv-payment-select tw-text-center">
+		<div
+			ref="container"
+			class="data-hj-suppress"
+			data-testid="braintree-drop-in"
+		></div>
+		<kv-loading-spinner
+			v-if="updatingPaymentWrapper"
+			class="tw-mb-2"
+		/>
+	</div>
+</template>
+
+<script lang="ts">
+import {
+	defineComponent, onMounted, ref, toRefs, watch,
+} from 'vue-demi';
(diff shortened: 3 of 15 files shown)


@kiva/kv-shop/src/useBraintreeDropIn.ts
```diff
@@ -0,0 +1,186 @@
...
+	const paymentMethodRequestable = ref(false);
+
+	function getApplePaymentRequest(amount: string): ApplePayPaymentRequest {
+		// @ts-ignore
+		return {
+			countryCode: 'US',
+			currencyCode: 'USD',
+			// merchantCapabilities: ['supports3DS'], // TODO: confirm/update
+			requiredBillingContactFields: ['postalAddress'],
+			// supportedNetworks: ['amex', 'discover', 'interac', 'jcb', 'masterCard', 'visa'], // TODO: confirm/update
```
emuvente, line 33: Which credit card networks do we actually accept? I know that we don't accept all the networks because we see errors like this crop up: "Invalid request: TRANSACTION_PAYMENT_METHOD_NONCE_CARD_TYPE_IS_NOT_ACCEPTED: Payment instrument type is not accepted by this merchant account."

I had to `@ts-ignore` several lines in the Dropin config because these are required values but somehow we've been using the Dropin so far without defining these options.

mcstover: Did this mark or come along with an update to the version of BT we're using?

In general, the interface had previously attempted to determine the card type based on input and we'd just let any errors on submission bubble up. I think Leo would know about the backend handling for exact payment types that we do or do not accept.

I guess if we "have" to start adding this config we can with the benefit it would likely provide more immediate feedback in the BT dropin interface to the user!

mcstover:
Really great start on this! Curious if you were able to link this and dev locally against CPS?

mcstover (approved):

@kiva/kv-shop/.eslintrc.cjs
```diff
@@ -0,0 +1,29 @@
...
+	extends: [
+		'../../.eslintrc.cjs',
+		'plugin:@typescript-eslint/eslint-recommended',
+		'plugin:@typescript-eslint/recommended',
+		'plugin:import/typescript',
+		'plugin:vue/recommended',
+		'@vue/typescript/recommended',
+	],
+	rules: {
+		'vue/html-indent': ['error', 'tab'],
```
dyersituations, line 19: Do we want to add these rules to the root `.eslintrc`?

@kiva/kv-shop/package.json
```diff
@@ -0,0 +1,53 @@
...
+        "eslint-plugin-vue": "^9.14.0",
+        "tsup": "^6.7.0",
+        "typescript": "^5.0.4",
+        "vue": "2.6",
+        "vue-eslint-parser": "^9.3.0"
+    },
+    "dependencies": {
+        "@apollo/client": "^3.7.14",
+        "@kiva/kv-components": "^3.19.0",
+        "@types/braintree-web-drop-in": "^1.34.2",
```
dyersituations, line 39: This can be a dev dep

@kiva/kv-shop/src/components/KvPaymentSelect.vue
```diff
@@ -0,0 +1,351 @@
...
+}
+@screen md {
+	.kv-payment-select [data-braintree-id="paypal-button"] {
+		width: 250px;
+	}
+}
+
+/* Hides credit card icon in number field until credit card type is known */
+.kv-payment-select [data-braintree-id="number-field-group"]:not(.braintree-form__field-group--card-type-known) svg {
+	display: none;
```
dyersituations, line 229: Some of these classes could be replaced with TW, but I assume this is all copied from the old repo.

@kiva/kv-shop/src/oneTimeCheckout.ts
```diff
@@ -0,0 +1,52 @@
...
+	apollo: any
+	transactionId: string
+}
+
+export async function waitOnTransaction({ apollo, transactionId }: WaitOnTransactionOptions) {
+	// TODO: get visitor ID
+	// poll transaction status and resolve when complete
+	const result = await apollo.query({
+		query: gql`
+			query checkoutStatus($transactionId: String!, $visitorId: string) {
```
dyersituations, line 39: I eventually want to get graphql file support (linting, codegen, etc.) in the other package, and we could reuse the pattern.

@kiva/kv-shop/src/shopError.ts
```diff
@@ -0,0 +1,45 @@
+export interface ShopErrorOptions {
```
dyersituations, line 1: Doesn't need to change, but in many cases `type` has replaced `interface` in TS. Kinda weird in my opinion, coming from C# world, but it's a general trend.

dyersituations:
> Really great start on this! Curious if you were able to link this and dev locally against CPS?

You might have seen this, but I added steps for linking in https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-loan-filters/README.md

@kiva/kv-shop/src/components/KvPaymentSelect.vue
emuvente: Yeah these were all copied from the old repo so I only really touched the styles that relied on foundation or sass in some way. Also `@apply` is transformed into the equivalent css for the tailwind classes used, which means there's no difference between `@apply tw-hidden;` and `display: none;`

dyersituations: For sure, and in some ways `display: none` is more readable...

@kiva/kv-shop/src/shopError.ts
dyersituations: For years `interface` was all the rage, which makes sense coming from Microsoft + C# land, but a few years ago they made `type` and `interface` essentially the same. I noticed that most 3rd party libraries used `type` as well, so I just assumed that behavior. I also worked with an engineer that I remember being super pro `type`... Haha. Anyways, here's the official explanation, basically there's one key difference around extensibility.

https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces

@kiva/kv-shop/src/components/KvPaymentSelect.vue
```diff
@@ -0,0 +1,351 @@
...
+
+/* Braintree Iframe is type number and inheriting some styles from kiva number input. */
+.kv-payment-select iframe[type=number] {
+	box-shadow: none;
+	padding: 0;
+	background-color: transparent;
+}
+
+/* TODO: Is this needed? */
+#payment-updating-overlay {
```
emuvente, line 240: @mcstover do you know what this is/was used for? I can't find any references to "payment-updating-overlay" in our current codebase or online.

@kiva/kv-shop/src/useBraintreeDropIn.ts
emuvente: The BT version in UI is 1.30.0, while the version here is 1.37.0, though I don't see anything in the changelogs about supportedNetworks being newly required.

The Dropin still initializes fine, so I think the options aren't actually required for the BT Dropin (the requirements come from the types as defined by Apple and Google, so BT might be applying default options when those aren't specified).

I think defining these options would be more like a UX enhancement.

@kiva/kv-shop/src/shopError.ts
emuvente: Aah, nice. I think that's the guide I read years ago (with updates). I've been following the rule of thumb described there:
> If you would like a heuristic, use `interface` until you need to use features from `type`.

@kiva/kv-shop/src/components/KvPaymentSelect.vue
eddieferrer: This can be removed. There used to be an overlay that got replaced by `kv-loading-spinner`

@kiva/kv-shop/package.json
emuvente: I tried that initially but ran into errors saying ti should be listed in the dependencies. Maybe that was a temporary issue though?

emuvente:
This one should be good to go now!

mcstover (approved):

christian14b (approved):

PR #249, opened 2023-05-25
fix: add core portion to gql import to match cps

christian14b:
This fix should come along with transpiling `kv-shop` on your project


mcstover (approved):
Worth a try

PR #251, opened 2023-05-26
fix: cookie basket sent encoded token

christian14b:


mcstover (approved):

mcstover (commented):
@christian14b can you please try add some tests to cover these utils so we can verify the logic before merging?

@kiva/kv-shop/src/tests/basket.spec.ts
```diff
@@ -0,0 +1,11 @@
+import { getCookieValue } from '../basket';
```
dyersituations, line 1: Doesn't need to change, but often the following happens in jest tests.

- A `describe` for each method, for example a second `describe('getCookieValue')`
- `it` strings typically start with "should", for example `it('should encode token')` for readability in test output

PR #252, opened 2023-05-26
fix: build command was targetting tests folder only

christian14b:


dyersituations (approved):
Looks like this should work. An alternative if this causes issues in the future is probably just:

```
tsup src/index.ts
```

Though I'm guessing the current solution allows choosing a single file when importing, so use with some caution.

PR #254, opened 2023-05-26
fix: always throwing error due to reading from undefined

christian14b:
Adding optional chaining on error checking

### DIFF
--- a/@kiva/kv-shop/src/basketItems.ts
+++ b/@kiva/kv-shop/src/basketItems.ts
@@ -32,11 +32,10 @@ export async function setTipDonation({ amount, apollo }: SetTipDonationOptions) 
 				basketId: getBasketID(),
 			},
 		});
-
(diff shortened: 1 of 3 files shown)


dyersituations (approved):

dyersituations:
@christian14b You should be able to test the entire process locally with `npm link`, rather than creating a lot of different releases.

mcstover (approved):
It may be that we need to handle an different error response signature here or extract the errors in a different way. Did you find an error response that didn't fit these handlers?

christian14b:
> It may be that we need to handle an different error response signature here or extract the errors in a different way. Did you find an error response that didn't fit these handlers?

@mcstover Not at all, parseShopError handles the error correctly, the problem was that we were trying to get the error from `result` when an error property didn't exist on success (Code like this condition: `result.error || result.errors.length`)

PR #257, opened 2023-06-02
Core 1372 location selector color changed

roger-in-kiva:
- color changed in KvLocationSelector component


@kiva/kv-components/vue/KvLocationSelector.vue
```diff
@@ -68,7 +68,7 @@
 		</fieldset>
 		<div class="tw-hidden md:tw-flex tw-gap-2 tw-justify-end tw-mt-1.5">
 			<button
-				class="tw-text-link"
+				class="tw-text-black"
```
dyersituations, line 71: Does not having a class result in the same expected text color?

roger-in-kiva: oh yeah, nice call, ty!

dyersituations (approved):

PR #261, opened 2023-06-15
feat: handle invalid basketId on new donation module

christian14b:
Check invalid basket id error, then create a new basket and save donation amount on cookie to be used on CPS.

### DIFF
--- a/@kiva/kv-shop/src/basket.ts
+++ b/@kiva/kv-shop/src/basket.ts
@@ -1,3 +1,6 @@
+import { gql } from '@apollo/client/core';
+import { parseShopError } from './shopError';
+
 // TODO: could be moved to shared file or separate package
 export const getCookieValue = (name: string) => {
 	if (typeof document !== undefined) {
@@ -10,6 +13,32 @@ export function getBasketID() {
 	return getCookieValue('kvbskt');
 }
 
-export function setBasketID(basketId: string) {
+export function setBasketID() {
 	// TODO
+}
+
+export async function createBasket(apollo) {
+	try {
+		apollo.mutate({
+			mutation: gql`mutation createNewBasketForUser { shop { id createBasket } }`,
+		}).then(({ data }) => {
+			const newBasketId = data.shop?.createBasket ?? null;
+			if (newBasketId) {
+				document.cookie = `kvbskt=${newBasketId}; path=/; secure=true;`;
+			}
+		});
+	} catch (error) {
+		throw parseShopError(error);
+	}
+}
+
+export function hasBasketExpired(errorCode) {
+	return ['shop.invalidBasketId', 'shop.basketRequired'].includes(errorCode);
+}
+
+export async function handleInvalidBasketForDonation({ donationAmount, navigateToCheckout = false, apollo }) {
+	document.cookie = `kvbskt=; expires=${new Date(0).toUTCString()}; path=/;`;
+	await createBasket(apollo);
+	document.cookie = `kvatbamt=${JSON.stringify({ donationAmount, navigateToCheckout })}; path=/;`;
+	window.location.reload();
 }
--- a/@kiva/kv-shop/src/basket.ts
+++ b/@kiva/kv-shop/src/basket.ts
@@ -37,8 +37,10 @@ export function hasBasketExpired(errorCode) {
 }
 
 export async function handleInvalidBasketForDonation({ donationAmount, navigateToCheckout = false, apollo }) {
-	document.cookie = `kvbskt=; expires=${new Date(0).toUTCString()}; path=/;`;
-	await createBasket(apollo);
-	document.cookie = `kvatbamt=${JSON.stringify({ donationAmount, navigateToCheckout })}; path=/;`;
-	window.location.reload();
+	if (typeof window !== 'undefined' && typeof document !== 'undefined') {
+		document.cookie = `kvbskt=; expires=${new Date(0).toUTCString()}; path=/;`;
+		await createBasket(apollo);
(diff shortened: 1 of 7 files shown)


@kiva/kv-shop/src/components/KvPaymentSelect.vue
```diff
@@ -22,6 +22,7 @@ import KvToast from '@kiva/kv-components/vue/KvToast.vue';
 import type { PropType } from 'vue-demi';
 import useBraintreeDropIn, { defaultPaymentTypes } from '../useBraintreeDropIn';
 import type { PayPalFlowType, PaymentType } from '../useBraintreeDropIn';
+import { getCookieValue } from '../basket';
```
mcstover, line 25: Seems unnecessary here...

@kiva/kv-shop/src/shopError.ts
```diff
@@ -35,6 +35,12 @@ export function parseShopError(error: any) {
 			original: error,
 		}, 'There was a problem validating your payment information. Please double-check the details and try again.');
 	}
+	if (errorCode === 'shop.invalidBasketId' || errorCode === 'shop.basketRequired') {
+		return new ShopError({
+			code: errorCode,
+			original: error,
+		}, 'There was a problem with your donation, refreshing the page to try again.');
```
mcstover, line 42: I don't think this is an appropriate message for all instances of these error codes. I think we should pass in a `context` param to this method that can be used to select context specific error messages.

@kiva/kv-shop/src/basket.ts
```diff
@@ -10,6 +13,32 @@ export function getBasketID() {
...
+		throw parseShopError(error);
+	}
+}
+
+export function hasBasketExpired(errorCode) {
+	return ['shop.invalidBasketId', 'shop.basketRequired'].includes(errorCode);
+}
+
+export async function handleInvalidBasketForDonation({ donationAmount, navigateToCheckout = false, apollo }) {
+	document.cookie = `kvbskt=; expires=${new Date(0).toUTCString()}; path=/;`;
```
mcstover, line 40: I think this would fail if used on the server due to the dependency on `document` and `window`.

We could either refactor to check for those before the code runs or maybe rename this to include "Client" in the name and add a doc block stating it's for clientside usage only.

@kiva/kv-shop/src/basketItems.ts
```diff
@@ -1,6 +1,6 @@
 import { gql } from '@apollo/client/core';
 import numeral from 'numeral';
-import { getBasketID } from './basket';
+import { getBasketID, hasBasketExpired, handleInvalidBasketForDonation } from './basket';
 import { parseShopError } from './shopError';
 
 export interface SetTipDonationOptions {
```
dyersituations, line 6: The `apollo` type can probably be `ApolloClient<any>`

@kiva/kv-shop/src/basketItems.ts
```diff
@@ -28,12 +30,29 @@ export async function setTipDonation({ amount, apollo }: SetTipDonationOptions)
...
 			}`,
 			variables: {
-				price: numeral(amount).format('0.00'),
+				price: donationAmount,
 				basketId: getBasketID(),
 			},
 		});
 		if (result?.error || result?.errors?.length) {
 			error = result?.error ?? result?.errors?.[0];
+			result?.errors?.forEach((err: any) => {
```
dyersituations, line 39: `any` should be something else here. If there isn't a built in type, you can create one custom in this file, based on the props we expect.

dyersituations: That could work. You could also inline it with `(result?.errors as GraphQLErrors ?? []).forEach`

christian14b: I was thinking the same, but I wasn't sure it was possible, thanks!

@kiva/kv-shop/src/basket.ts
```diff
@@ -10,6 +13,34 @@ export function getBasketID() {
...
+	} catch (error) {
+		throw parseShopError(error);
+	}
+}
+
+export function hasBasketExpired(errorCode) {
+	return ['shop.invalidBasketId', 'shop.basketRequired'].includes(errorCode);
+}
+
+export async function handleInvalidBasketForDonation({ donationAmount, navigateToCheckout = false, apollo }) {
```
dyersituations, line 39: While it won't be trivial, long-term we'll want unit tests on these new methods. Please at least add a ticket to come back and address.

dyersituations: Same with the `basketItems` file. Looks like that file has no tests.

@kiva/kv-shop/src/basketItems.ts
```diff
@@ -1,16 +1,19 @@
...
 import numeral from 'numeral';
-import { getBasketID } from './basket';
+import { GraphQLErrors } from '@apollo/client/errors';
+import { getBasketID, hasBasketExpired, handleInvalidBasketForDonation } from './basket';
 import { parseShopError } from './shopError';
 
 export interface SetTipDonationOptions {
 	amount: string | number,
-	apollo: any,
+	apollo: ApolloClient<any>,
```
dyersituations, line 9: Actually, this could be `ApolloClient<NormalizedCacheObject>` to get around the linting error.

@kiva/kv-shop/src/basket.ts
christian14b: Added some basic tests but left a comment to add remaining cases

dyersituations (approved):

mcstover (approved):

PR #269, opened 2023-06-27
feat: disabling to click loan use when loading

roger-in-kiva:
- class added to loan use link to disable clicking when loading


@kiva/kv-components/vue/KvClassicLoanCard.vue
```diff
@@ -100,6 +100,7 @@
 					:to="readMorePath"
 					:href="readMorePath"
 					class="loan-card-use tw-text-primary"
+					:class="{ 'tw-pointer-events-none' : isLoading }"
```
dyersituations, line 103: Would it be better to put this on the entire loan card?

dyersituations (approved):

PR #271, opened 2023-07-14
feat: custom countdown for vue 2 + 3

dyersituations:
https://kiva.atlassian.net/browse/CORE-928

- The countdown component we used had vue 2 or 3 versions, not one that worked in either
- Here's a simple alternative using `moment` - it looks identical
- The package lock already included `moment`, so this isn't introducing a new library
- Removed logging we no longer need

image

### DIFF
--- a/@kiva/kv-components/package.json
+++ b/@kiva/kv-components/package.json
@@ -53,7 +53,6 @@
     "build": "echo No build needed for @kiva/kv-components."
   },
   "dependencies": {
-    "@chenfengyuan/vue-countdown": "^1.1.5",
     "@kiva/kv-tokens": "^2.8.0",
     "@mdi/js": "^5.9.55",
     "@vueuse/integrations": "^7.6.0",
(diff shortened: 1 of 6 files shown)


eddieferrer:
Nice. If the countdown was its own component we could port this back into UI and remove the vue-countdown package there as well.

roger-in-kiva (approved):
Looks good!
Is there a way to have these components in vue3?

dyersituations:
> Looks good! Is there a way to have these components in vue3?

What do you mean? The components will work in vue3.

roger-in-kiva:
> > Looks good! Is there a way to have these components in vue3?
> 
> 
> 
> What do you mean? The components will work in vue3.

Surely they'll work!
I mean using vue-demi and the setup hook, like others...

dyersituations:
> Surely they'll work! I mean using vue-demi and the setup hook, like others...

Not necessary, the loan card components are also in this older format. But I'll use that format for a extracted timer.

dyersituations:
@roger-in-kiva @eddieferrer I've reopened the PR with an abstracted out countdown component. My original thought was that this component was only needed in the loan card, but I see now that the borrower profile will also need it long-term.

mcstover (approved):

eddieferrer (commented):
Excellent! 👍

PR #274, opened 2023-07-26
feat: 5 dollars pre select added to loan card components

roger-in-kiva:
- 5 dollars pre-select dropdown option added to loan cards components

### DIFF
--- a/@kiva/kv-components/utils/loanUtils.js
+++ b/@kiva/kv-components/utils/loanUtils.js
@@ -77,10 +77,6 @@ export function getLendCtaSelectedOption(
 			// Top up campaign defaults to $5
 			return Number(unreservedAmount > 5 ? 5 : unreservedAmount).toFixed();
 		}
-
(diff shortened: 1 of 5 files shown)


@kiva/kv-components/utils/loanUtils.js
```diff
@@ -76,6 +77,14 @@ export function getLendCtaSelectedOption(
 			// Top up campaign defaults to $5
 			return Number(unreservedAmount > 5 ? 5 : unreservedAmount).toFixed();
 		}
+
+		if (fiveDollarsSelected) {
```
dyersituations, line 81: This isn't needed and will be covered by the following return statement.

dyersituations (approved):

@kiva/kv-components/utils/loanUtils.js
roger-in-kiva: That's true, thanks!

PR #276, opened 2023-08-01
feat: cleanup unneeded code in kiva partner header

dyersituations:
https://kiva.atlassian.net/browse/CIT-229

- The `auth0` repo didn't like these (mainly the empty `setup`)

### DIFF
--- a/@kiva/kv-components/vue/KvKivaPartnerHeader.vue
+++ b/@kiva/kv-components/vue/KvKivaPartnerHeader.vue
@@ -32,13 +32,7 @@
 	</nav>
 </template>
 
-<script>
-export default {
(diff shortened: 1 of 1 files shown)


mcstover (approved):
I think importing the vue-demi methods would have made this compatible with the vue 2 context. But for now, there's no js needed so clearing it out should be fine too.

dyersituations:
> I think importing the vue-demi methods would have made this compatible with the vue 2 context. But for now, there's no js needed so clearing it out should be fine too.

Yeah the other components from kv-components work fine, but some reason the compilation in `auth0` freaked out with the empty `setup`.

PR #277, opened 2023-08-02
feat: balance campaign prices added to loan utility

roger-in-kiva:
- lower default price for lenders coming from repayments email added to loan utility
- requirements here: https://kiva.atlassian.net/browse/CORE-1468

### DIFF
--- a/@kiva/kv-components/utils/loanUtils.js
+++ b/@kiva/kv-components/utils/loanUtils.js
@@ -1,6 +1,16 @@
 export const ERL_COOKIE_NAME = 'kverlfivedollarnotes';
 export const TOP_UP_CAMPAIGN = 'TOPUP-VB-BALANCE-MPV1';
 export const BASE_CAMPAIGN = 'BASE-VB_BALANCE_MPV1';
+export const BALANCE_CAMPAIGN = 'REPAYMENT-NOTIFICATION_BALANCE_MPV1';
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/utils/loanUtils.js
```diff
@@ -56,33 +71,36 @@ export function getLendCtaSelectedOption(
 			// eslint-disable-next-line no-nested-ternary
 			currentCampaign = campaignToCheck.includes(TOP_UP_CAMPAIGN)
 				? TOP_UP_CAMPAIGN
-				: (campaignToCheck.includes(BASE_CAMPAIGN) ? BASE_CAMPAIGN : '');
+				: campaignToCheck.includes(BASE_CAMPAIGN) ? BASE_CAMPAIGN // eslint-disable-line no-nested-ternary
+					: campaignToCheck.includes(BALANCE_CAMPAIGN) ? BALANCE_CAMPAIGN // eslint-disable-line no-nested-ternary
+						: (campaignToCheck.includes(NO_BALANCE_CAMPAIGN) ? NO_BALANCE_CAMPAIGN : '');
```
mcstover, line 76: This is getting pretty messy. We'll likely want to find a better way if these campaign types keep expanding...

mcstover (approved):

@kiva/kv-components/utils/loanUtils.js
roger-in-kiva: Agreed!
Only issue I see is that we have different conditions for every campaign. Could we get them from kiva admin?

PR #281, opened 2023-08-14
feat: avoid fractional page number in kvpaginator component

roger-in-kiva:
- avoid fractional page number in `KvPaginator` component

### DIFF
--- a/@kiva/kv-components/vue/KvPagination.vue
+++ b/@kiva/kv-components/vue/KvPagination.vue
@@ -109,7 +109,7 @@ export default {
 	},
 	computed: {
 		current() {
-			const page = this.offset / this.limit;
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/vue/KvPagination.vue
```diff
@@ -109,7 +109,7 @@ export default {
 	},
 	computed: {
 		current() {
-			const page = this.offset / this.limit;
+			const page = Math.floor(this.offset / this.limit) + 1;
```
dyersituations, line 112: Would this be the same as Math.ceil?

roger-in-kiva: yeah, and the best to pass the test, ty!

PR #286, opened 2023-08-29
feat: kvmap ported to kv components library

roger-in-kiva:
- kvmap component ported from ui to kv-components library
- leaflet and maplibre-gl libraries updated to last version
- keep it on vue2 to test it in cps repo


@kiva/kv-components/vue/KvMap.vue
```diff
@@ -0,0 +1,327 @@
...
+			:id="`kv-map-holder-${mapId}`"
+			:ref="refString"
+			class="tw-w-full tw-h-full tw-bg-black"
+			:style="{ position: 'absolute' }"
+		></div>
+	</div>
+</template>
+
+<script>
+import { createIntersectionObserver } from '../utils/observerUtils';
```
dyersituations, line 16: Since we don't have existing utils in this project, you can just copy in these two methods (`createIntersectionObserver` and `testDelayedGlobalLibrary`) to this component for now. They aren't used elsewhere in this library as well.

@kiva/kv-components/package-lock.json
dyersituations, line 1: We don't want this lock file. Instead the lock file in root should be updated.

roger-in-kiva: Gotcha! How to fix fails below?

dyersituations (approved):

PR #287, opened 2023-08-30
fix: kvmap component refactor to avoid using vue meta package

roger-in-kiva:
- kvmap component refactor to avoid using vue meta package and having a shareable component

### DIFF
--- a/@kiva/kv-components/vue/KvMap.vue
+++ b/@kiva/kv-components/vue/KvMap.vue
@@ -15,43 +15,6 @@
 <script>
 export default {
 	name: 'KvMap',
-	metaInfo() {
-		return {
-			script: [].concat(!this.hasWebGL ? [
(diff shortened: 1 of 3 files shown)


@kiva/kv-components/vue/KvMap.vue
```diff
@@ -168,6 +131,7 @@ export default {
 		if (!this.mapLibreReady && !this.leafletReady) {
 			this.initializeMap();
 		}
+		this.getLibraries();
```
mcstover, line 134: I think this call can go alongside `initializeMap` within the check above. We don't want to load the libraries multiple times...

We might also consider calling it within each block of the initializeMap method which will have established webGL support status by then and Also start loading the libarries before the checks start.

roger-in-kiva: Nice call, thanks!

dyersituations (approved):

PR #288, opened 2023-08-31
feat: add box shadows to tokens

dyersituations:
- Used the default TW shadow for the default
- Added the shadow we used for Kiva cards as the `md` variation, since it's a fairly wide shadow

default:
image

`md`:
image


mcstover (approved):

eddieferrer:
Cool! This does look awfully similar to `tw-drop-shadow-lg`

dyersituations:
> Cool! This does look awfully similar to `tw-drop-shadow-lg`

I suppose we could just roll with the TW defaults.

dyersituations:
The TW shadows are a decent amount darker, but the size of our loan card shadow is closer to the default `lg` class, so I renamed it to `lg`, but I retained the lighter shadow.

PR #289, opened 2023-09-01
Toast variations

dyersituations:
https://kiva.atlassian.net/browse/CIT-328

- Added TW configs for box shadow
- Added bg variations
- Added flag for adding drop/box shadow

image

image

image

image

### DIFF
--- a/@kiva/kv-components/vue/KvToast.vue
+++ b/@kiva/kv-components/vue/KvToast.vue
@@ -21,10 +21,10 @@
 						class="
 							tw-rounded tw-overflow-hidden
 							tw-flex
-							tw-bg-secondary
 							tw-mx-auto
 							tw-w-full md:tw-w-max md:tw-max-w-full md:tw-min-w-1/2
 						"
+						:class="[backgroundClass, { 'tw-shadow': applyDropShadow }]"
 					>
 						<div
 							class="
@@ -87,13 +87,13 @@
 
 						<button
 							class="
-							tw-w-5
-							tw-flex-shrink-0
-							tw-flex
-							tw-items-center tw-justify-center
-							tw-bg-secondary
-							hover:tw-text-action-highlight
-						"
+								tw-w-5
+								tw-flex-shrink-0
+								tw-flex
+								tw-items-center tw-justify-center
+								hover:tw-text-action-highlight
+							"
+							:class="[backgroundClass]"
 							@click="close"
 						>
 							<kv-material-icon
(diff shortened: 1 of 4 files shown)


mcstover (approved):

@kiva/kv-components/vue/KvToast.vue
```diff
@@ -21,10 +21,10 @@
 						class="
 							tw-rounded tw-overflow-hidden
 							tw-flex
-							tw-bg-secondary
 							tw-mx-auto
 							tw-w-full md:tw-w-max md:tw-max-w-full md:tw-min-w-1/2
 						"
+						:class="[backgroundClass, { 'tw-shadow': applyDropShadow }]"
```
emuvente, line 27: Why not always apply the drop shadow? I think it would be weird for it to sometimes have it and sometimes not. It's also not going to break anything to make that the new default. 

I'm also dubious about being able to pick the background style, but I don't know what problem is trying to be solved there. I would think that we would just update the toast style if the current style wasn't working.

dyersituations: Good call, and fine by me. I was cautious to affect the existing defaults. That being said, I'm out the next few days, but I'll make a note to propose a new default to design late next week! Thanks for taking a look at the changes.

mcstover: A background color change was requested b/c we're using tw-secondary as the page background and toast messages were "disappearing" when shown over top of it. With that in mind, I suggested providing the 3 basic options we also provide for page background color in contentful.

It's true that the drop shadow begins to solve the issue on it's own but the toast is still pretty muted when it matches the background of the page. Of course this is the same problem we've had for a while in other contexts and this will add complexity to the Tip message logic if we want to pass a param through that calling function.

emuvente: What page specifically is having issues? Most pages use the `primary` white background, so I would first look at changing the background color of the page in question. If that's really not an option, then we should adjust the default styles of the toast so that it works with whichever background is behind it. I don't think we should add complexity and inconsistency.

I get the idea of changing the toast background based on the page background parameter, but even that won't always work because many pages have multiple background colors as you scroll down the page. There will always be some part of some page where the toast bg will be the same as the bg of the content behind it.

Since there is no perfect solution, I would rather go with the imperfect solution that doesn't add complexity and just change the default styles to something that works well enough.

mcstover: The new impact-dashboard uses the secondary bg color coupled with a cards type interface where the cards have a white bg. And ultimately, there are many many more bg colors at play now. This request came from design so we can take it back to them and try to work out an option that comes up with a better default that will work in "most" contexts.

We didn't change any defaults with this implementation and I'll note that we shouldn't implement or use the variation on the background color in our current work.

PR #291, opened 2023-09-07
fix: retry setting tip donation after creating new basket MARS-429

emuvente:
The invalid basket errors were still coming through snowplow and were presumably displayed to the user before the page was refreshed. This switches to retrying the set tip donation mutation without refreshing the page, up to 2 times.

### DIFF
--- a/@kiva/kv-shop/src/basketItems.ts
+++ b/@kiva/kv-shop/src/basketItems.ts
@@ -1,70 +1,76 @@
-import { gql, ApolloClient, NormalizedCacheObject } from '@apollo/client/core';
+import { gql, ApolloClient, DocumentNode } from '@apollo/client/core';
 import numeral from 'numeral';
-import { GraphQLErrors } from '@apollo/client/errors';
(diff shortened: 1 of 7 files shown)


mcstover (approved):
Nice refactor

@kiva/kv-shop/src/basketItems.ts
```diff
@@ -1,70 +1,76 @@
...
-	let error;
-	let hasFailedAddToBasket = false;
-	const donationAmount = numeral(amount).format('0.00');
+/**
+ * Call a shop mutation with a basketId, creating a new basket if necessary.
+ */
+async function callShopMutation(
+	apollo: ApolloClient<any>,
+	mutation: DocumentNode,
+	variables: Record<string, any>,
```
dyersituations, line 12: GitHub linting is catching our usages of `any`. While not a terrible thing to catch, we could consider turning off that linting requirement to reduce noise in situations like `Record<string, any>` where it isn't as important to have a specific type.

dyersituations (approved):

PR #295, opened 2023-09-13
feat: new edit button that opens up a lightbox for editing page setti…

michelleinez:
…ngs in kiva at work pages

### DIFF
--- a/@kiva/kv-components/vue/KvEditButton.vue
+++ b/@kiva/kv-components/vue/KvEditButton.vue
@@ -12,8 +12,10 @@
 
 		<!-- Lightbox -->
 		<KvLightbox
-			:visible="isLightboxVisible"
-			:title="lightboxTitle"
+			:visible="visible"
+			:title="title"
+			:prevent-close="preventClose"
+			:variant="variant"
 			@lightbox-closed="hideLightbox"
 		>
 			<slot></slot>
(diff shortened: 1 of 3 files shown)


@kiva/kv-components/vue/KvEditButton.vue
```diff
@@ -0,0 +1,57 @@
...
+	},
+	data() {
+		return {
+			isLightboxVisible: false,
+			mdiPencil,
+		};
+	},
+	methods: {
+		showLightbox() {
+			this.isLightboxVisible = true;
```
mcstover, line 50: Can you please emit and event from both this `showLightbox` and the `hideLightbox` methods? This will allow us to listen for that event in the parent and fire analytics events according to whatever context we're working in.

eddieferrer:
A "nice to have" would be the ability to support adding content to the controls and header slot in the lightbox in addition to the default slot.

@kiva/kv-components/vue/KvEditButton.vue
michelleinez: done

@kiva/kv-components/vue/KvEditButton.vue
```diff
@@ -0,0 +1,92 @@
...
+			default: false,
+		},
+	},
+	emits: [
+		'lightbox-opened',
+		'lightbox-closed',
+	],
+	setup({ emit }) {
+		const showLightbox = () => {
+			this.visible = true;
```
dyersituations, line 73: `visible` needs to be a ref, here `this` is undefined.

michelleinez: I'm just going to make this component manage state instead of the parent component.
I think it makes sense for the lightbox component itself to have visibility set by a parent component, but for the edit button, the button controls the visibility of the lightbox and that makes integrating the edit button simpler.

If we happen to run into a need to have the lightbox be open by default we can always add support for that afterwards..

mcstover (approved):
Nice combo of components and will cut out a lot of boilerplate in our upcoming work!

@kiva/kv-components/vue/stories/KvEditButton.stories.js
```diff
@@ -0,0 +1,64 @@
...
+		preventClose: false,
+	},
+};
+
+const DefaultTemplate = (args, { argTypes }) => ({
+	props: Object.keys(argTypes),
+	components: { KvEditButton, KvButton },
+	template: `
+		<div>
+			<p>The lightbox is visible: {{isLightboxVisible}}</p>
```
dyersituations, line 39: This value doesn't currently change when the lightbox is opened. Maybe it isn't supposed to?

dyersituations:
Mocks show the button as square. Do we want to adjust that now?

dyersituations (approved):

mcstover (approved):

PR #296, opened 2023-09-14
Cit 334 line graph

dyersituations:
https://kiva.atlassian.net/browse/CIT-334

- Created line graph component using the stash from @emuvente as the base
- Added support for labels (value and x-axis)
- Added ability to just pass in an array of values w/ labels
- Labels will likely get cut-off (or overflow really) in many cases, but this is a good first pass for the impact dashboard
- Line isn't perfect (gets thinner with larger slope, but again, okay for first pass)
- Added some basic tests

image

image

image

image


mcstover (approved):
Neat!

PR #297, opened 2023-09-16
fix: backpedaling to reduce edit button functionality

michelleinez:


@kiva/kv-components/vue/KvEditButton.vue
```diff
@@ -1,104 +1,30 @@
...
-			<template #controls>
-				<slot
-					name="controls"
-					:hideLightbox="hideLightbox"
-				></slot>
-			</template>
-		</KvLightbox>
-	</div>
+	<!-- Edit Button -->
+	<KvButton
```
dyersituations, line 3: Will this component get more functionality in the near future? If not, I'm not sure we need this as a shared component in our UI library. It's currently just a button with an icon.

michelleinez: Okay, no storybook component needed for this then.

PR #298, opened 2023-09-19
feat: increase impact visibility for lse loans

roger-in-kiva:
- Increasing impact visibility for LSE loans with next features:

1. Add a loan tag for LSE loans that says “High community impact”  
2. Change priority of grey “pills” so that we only show one public tag.
      Priority order: 1.Tag, 2. Category, 3. Activity, 4. Sector, 5. Theme

- For now we’ll use the lending partner containing “N/A” as the signifier for these loans

### DIFF
--- a/@kiva/kv-components/vue/KvClassicLoanCard.vue
+++ b/@kiva/kv-components/vue/KvClassicLoanCard.vue
@@ -231,6 +231,8 @@ import KvLoanTag from './KvLoanTag.vue';
 import KvMaterialIcon from './KvMaterialIcon.vue';
 import KvLoadingPlaceholder from './KvLoadingPlaceholder.vue';
 
+const LSE_LOAN_KEY = 'N/A';
+
 export default {
 	name: 'KvClassicLoanCard',
 	components: {
@@ -438,8 +440,19 @@ export default {
 				singleParents: !!tags.filter((t) => t.toUpperCase() === 'SINGLE PARENT').length,
 			};
 
+			const isLseLoan = this.loan?.partnerName?.includes(LSE_LOAN_KEY);
+
 			// P1 Category
 			// Exp limited to: Eco-friendly, Refugees and IDPs, Single Parents
+			// Tag as first option for LSE loans
+			if (isLseLoan) {
+				const position = Math.floor(Math.random() * tags.length);
+				const tag = tags[position];
+				if (!callouts.filter((c) => c.toUpperCase() === tag.toUpperCase()).length) {
+					callouts.push(tag);
+				}
+			}
+
 			if (!this.categoryPageName) {
 				if (categories.ecoFriendly) {
 					callouts.push('Eco-friendly');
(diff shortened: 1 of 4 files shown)


@kiva/kv-components/vue/KvClassicLoanCard.vue
```diff
@@ -438,8 +440,19 @@ export default {
 				singleParents: !!tags.filter((t) => t.toUpperCase() === 'SINGLE PARENT').length,
 			};
 
+			const isLseLoan = this.loan?.partnerName?.includes(LSE_LOAN_KEY);
```
dyersituations, line 443: We probably want to add a `toUpperCase` to cover casing differences.

@kiva/kv-components/vue/KvLoanTag.vue
```diff
@@ -55,6 +57,10 @@ export default {
 			return null;
 		},
 		tagText() {
+			const partnerName = this.loan?.partnerName ?? '';
+			if (partnerName.includes(LSE_LOAN_KEY)) {
```
dyersituations, line 61: Same comment about `toUpperCase`

@kiva/kv-components/vue/KvLoanTag.vue
```diff
@@ -55,6 +57,10 @@ export default {
...
+				return 'High community impact';
```
dyersituations, line 62: So we're showing both a custom tag and callout for the LSE loans? Do you have a screenshot example from our dev data?

@kiva/kv-components/vue/KvClassicLoanCard.vue
```diff
@@ -438,8 +440,19 @@ export default {
...
+
 			// P1 Category
 			// Exp limited to: Eco-friendly, Refugees and IDPs, Single Parents
+			// Tag as first option for LSE loans
+			if (isLseLoan) {
+				const position = Math.floor(Math.random() * tags.length);
```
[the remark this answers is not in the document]
roger-in-kiva: Just replicating from here: https://github.com/kiva/kv-ui-elements/blob/023c2d950460478d6c3684d3d09906f9d9672f35/%40kiva/kv-components/vue/KvClassicLoanCard.vue#L468

I suppose to show one of the tags are in the array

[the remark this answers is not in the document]
dyersituations: In that case, do we need to instead find the first tag that doesn't exist in the callouts? It looks like we're currently pulling a random tag, and not adding it if it matches. What if there's another tag that isn't in callouts?

@kiva/kv-components/vue/KvLoanTag.vue
roger-in-kiva: We are just showing a tag for LSE loans.
For callouts we are just changing the priority order to have the tag first if exists and just showing one callout.
This is the priority order for LSE loans 1.Tag, 2. Category, 3. Activity, 4. Sector, 5. Theme instead of 1. Category, 2. Activity, 3. Sector, 4. Tag, 5. Theme. for others

dyersituations (approved):

PR #302, opened 2023-09-28
Extend loan use with why special

dyersituations:
- Added option why special string to loan use so we can re-use the component in impact dashboard

image


roger-in-kiva (approved):
Nice!

eddieferrer (approved):

mcstover (approved):

PR #304, opened 2023-10-04
fix: bug fixed for lse loans when including eco friendly and single p…

roger-in-kiva:
- LSE loans bug fixed when including eco friendly and single parent tags

### DIFF
--- a/@kiva/kv-components/vue/KvClassicLoanCard.vue
+++ b/@kiva/kv-components/vue/KvClassicLoanCard.vue
@@ -456,13 +456,13 @@ export default {
 			}
 
 			if (!this.categoryPageName) {
-				if (categories.ecoFriendly
-					&& !callouts.find((c) => c === ECO_FRIENDLY_KEY || c === SUSTAINABLE_AG_KEY)) {
+				if (categories.ecoFriendly // eslint-disable-next-line max-len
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/vue/KvClassicLoanCard.vue
```diff
@@ -453,12 +457,12 @@ export default {
 
 			if (!this.categoryPageName) {
 				if (categories.ecoFriendly
-					&& !callouts.filter((c) => c.toUpperCase() === categories.ecoFriendly.toUpperCase()).length) {
+					&& !callouts.find((c) => c === ECO_FRIENDLY_KEY || c === SUSTAINABLE_AG_KEY)) {
```
dyersituations, line 460: Do we still need case insensitive checking here?

roger-in-kiva: Looks like test is passing, but good for consistency. ty

dyersituations:
The changes don't look that different from what was there previously. What specifically caused the issue?

roger-in-kiva:
> The changes don't look that different from what was there previously. What specifically caused the issue?

Previously we were comparing with this `categories.ecoFriendly.toUpperCase()` but `categories.ecoFriendly` is boolean.
Now we are checking if keys exist in the callouts array.

dyersituations (approved):

PR #313, opened 2023-11-10
feat: one-time checkout and supporting functions MARS-546

emuvente:
One-time async checkout, with some basket state and validation functions. Still needs to actually be tried out with an app, more tests, and a solution for tracking.

Despite the roughness, the basic flow should be all set here. Please let me know if anything looks off with the flow as it is now or if there are any missing pieces not already called out with a TODO!

### DIFF
--- a/@kiva/kv-shop/src/useBraintreeDropIn.ts
+++ b/@kiva/kv-shop/src/useBraintreeDropIn.ts
@@ -1,9 +1,10 @@
 /* eslint-disable @typescript-eslint/ban-ts-comment */
 import { gql } from '@apollo/client/core';
 import numeral from 'numeral';
+import type { Ref } from 'vue-demi';
 import { ref } from 'vue-demi';
 import type { ApplePayPaymentRequest } from 'braintree-web';
-import type { Dropin } from 'braintree-web-drop-in';
+import type { Dropin, PaymentMethodPayload } from 'braintree-web-drop-in';
 import { ShopError, parseShopError } from './shopError';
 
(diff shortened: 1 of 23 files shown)


mcstover (approved):
Great advancement of kv-shop methods and utils!

mcstover (approved):

@kiva/kv-shop/src/useBraintreeDropIn.ts
```diff
@@ -44,7 +45,14 @@ export interface DropInInitOptions {
...
 
-export default function useBraintreeDropIn() {
+export interface DropInWrapper {
+	initDropIn: (options: DropInInitOptions) => Promise<Dropin>,
+	paymentMethodRequestable: Ref<boolean>,
+	requestPaymentMethod: () => Promise<PaymentMethodPayload | void>,
+	updateAmount: (amount: string|number) => void,
+}
+
+function initBraintreeDropin(): DropInWrapper {
```
dyersituations, line 55: I can't comment on the line, but GitHub linting is saying that `event` is never used on line 95.

dyersituations (approved):

@kiva/kv-shop/src/useBraintreeDropIn.ts
emuvente: We do have a `// eslint-disable-next-line no-unused-vars` comment on it, but maybe github doesn't evaluate those in-line linting rules?

dyersituations: Not sure, haven't spent much time with GitHub's linter.

PR #309, opened 2023-10-26
feat: initial www header MARS-528

emuvente:
New site navigation header for www.

Remaining work:

- [ ] visual review
- [ ] search suggestions
- [ ] lend menu data
- [ ] trustee/borrower states
- [ ] tracking events
- [ ] interface docs

PR #314, opened 2023-11-13
feat: load and format lend menu data

christian14b:


@kiva/kv-components/package.json
```diff
@@ -63,6 +63,8 @@
     "embla-carousel": "^4.5.3",
     "focus-trap": "^6.7.2",
     "fuse.js": "^6.6.2",
+    "lodash.groupby": "^4.6.0",
```
emuvente, line 66: Instead of using groupby, could you please refactor [this function](https://github.com/kiva/kv-ui-elements/pull/309/files#diff-8401c4f3d2b2f656dcb16c9dda8b1aa1d3ed72de71117384bd59cb72024f38bbR39-R48) into a generic util and use that instead?

@kiva/kv-components/package.json
```diff
@@ -63,6 +63,8 @@
...
+    "lodash.sortby": "^4.7.0",
```
emuvente, line 67: Instead of sortBy, let's make our own util function to do this https://github.com/you-dont-need/You-Dont-Need-Lodash-Underscore#_sortby-and-_orderby

@kiva/kv-components/utils/comparators.js
```diff
@@ -38,3 +38,26 @@ export function startsWith(query, key) {
...
+ *
+ * const list = ['a', 'b', 'c', 'd'];
+ * const arrayToSort = ['d', 'a'];
+ * arrayToSort.sort(indexIn(list)); // will return ['a','d']
+ *
+ * @param {array} list
+ * @param {string} [key] - compare property 'key' of elements rather than the elements themselves
+ * @returns {function}
+ */
+export function indexIn(list, key) {
```
emuvente, line 54: Thanks for adding this!

@kiva/kv-components/vue/KvWwwHeader.vue
```diff
@@ -142,14 +147,31 @@ export default {
...
+	setup(props, { emit }) {
 		const linksVisible = ref(true);
 		const searchBar = ref(null);
 		const searchOpen = ref(false);
 		const activeHeaderItem = ref(null);
 		const menuOpen = ref(false);
-		const menuComponent = ref(null);
+		const menuComponent = shallowRef(null);
+		const menuComponentInstance = ref(null);
+		const userId = ref(props.userId);
```
emuvente, line 174: userId should be declared as a prop, and then it will be available in the template and won't need to be declared here.

emuvente (commented):
Looking good! Just one thing to clear up

@kiva/kv-components/vue/KvWwwHeader.vue
dyersituations: Follow-up here, can this const be removed entirely if it's a prop?

PR #319, opened 2023-11-21
feat(KvPaymentSelect): emit error instead of using broken local toast

emuvente:
Ran into errors of errorToast being undefined. To fix, I switched to emitting the error message so that consuming apps can display the error (or not) in whatever way they want.

### DIFF
--- a/@kiva/kv-shop/src/components/KvPaymentSelect.vue
+++ b/@kiva/kv-shop/src/components/KvPaymentSelect.vue
@@ -9,7 +9,6 @@
 			v-if="updatingPaymentWrapper"
 			class="tw-mb-2"
 		/>
-		<kv-toast :ref="errorToast" />
 	</div>
 </template>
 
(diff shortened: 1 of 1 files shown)


mcstover (approved):
I think this is fine. We may also want to consider moving this component out to kv-components instead to reduce the need for special configs to resolve components out of the kv-shop `dist` folder. I made progress getting KvPaymentSelect to resolve in our CPS tests but it just leads to a different issue.

emuvente:
> We may also want to consider moving this component out to kv-components instead to reduce the need for special configs to resolve components out of the kv-shop dist folder.

KvPaymentSelect depends on the `useBraintreeDropIn` composable, which much of the rest of kv-shop also uses. That means moving KvPaymentSelect to kv-components would require adding kv-shop as a dependency to kv-components. That didn't feel like the right order of dependencies to me.

> I made progress getting KvPaymentSelect to resolve in our CPS tests but it just leads to a different issue.

I would be curious to know what the different issues you ran into after the tests passed. I think the solution to those kind of issues should be resolved with adjusting the publishing process or adjusting the consumer build process, rather than moving components around.

mcstover:
@emuvente I didn't think about that deeper dependency there. Basically once, KvPaymentSelect started to resovle properly, by specifically adding concessions for `@kiva/kv-shop/dist` the other issues started popping up.

PR #320, opened 2023-11-28
Horizontal loan card

eddieferrer:
New horizontal loan card and a couple of bug fixes:

* Add horizontal loan card
* Support square images in KvBorrowerImage
* Change 'unreservedAmount' to be a number, stringify to string in `kv-loan-progress-group` prop
* Add the ability to add custom callouts to both loan cards via the `customCallouts` prop

Screenshot 2023-11-28 at 12 08 29 PM


roger-in-kiva (approved):
Nice work, thanks for the mixin.
Should we have a limit for custom callouts?

@kiva/kv-components/tests/unit/specs/components/KvWideLoanCard.spec.js
```diff
@@ -0,0 +1,63 @@
...
+						fundraisingPercent: undefined,
+					},
+					kvTrackFunction,
+					photoPath,
+					externalLinks: true,
+				},
+			});
+		const results = await axe(container);
+		expect(results).toHaveNoViolations();
+	});
```
mcstover, line 62: Coule we add an additionall test or 2 that checks the results of your loanCard util? If we can test the rendering of a couple of those computed properties across vue2 and vue3 that would help verify the functionality.

mcstover (approved):
Thanks for the additional tests!

@kiva/kv-components/tests/unit/specs/components/KvWideLoanCard.spec.js
eddieferrer: Added tests of the country display. Since that seem pretty useful. It tests `formattedLocation` `countryName` and `distributionModel` computed props

PR #322, opened 2023-11-30
kv-analytics library

emuvente:
I need this for tracking donation transaction in optimizely in kv-shop. Largely just copied from UI. This allows these common tracking functions to be used across kv-ui-elements and in UI/CPS.


mcstover (approved):
Nice!

PR #326, opened 2023-12-06
fix(KvLightbox): allow focus on open alerts, clicks on browser extensions and outside lightbox

emuvente:
The focus trap was preventing closing tip message toasts and preventing clicking outside the lightbox to close it.

Using the `allowOutsideClick: true` option allows clicking outside the focus trap, which then allows closing toasts and accessing elements added by browser extensions like password managers.

Including the `'[role="alert"]'` selector string in the focus trap array will include any open alerts in the focus trap, allowing for focusing on the alert close button using tab.

### DIFF
--- a/@kiva/kv-components/vue/KvLightbox.vue
+++ b/@kiva/kv-components/vue/KvLightbox.vue
@@ -218,10 +218,16 @@ export default {
 		const kvLightboxBody = ref(null);
 		const controlsRef = ref(null);
 
+		const trapElements = computed(() => [
+			kvLightbox.value, // This lightbox
+			'[role="alert"]', // Any open toasts/alerts on the page
+		]);
 		const {
 			activate: activateFocusTrap,
 			deactivate: deactivateFocusTrap,
-		} = useFocusTrap(kvLightbox);
+		} = useFocusTrap(trapElements, {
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/vue/KvLightbox.vue
```diff
@@ -218,10 +218,16 @@ export default {
...
+		const trapElements = computed(() => [
+			kvLightbox.value, // This lightbox
+			'[role="alert"]', // Any open toasts/alerts on the page
+		]);
 		const {
 			activate: activateFocusTrap,
 			deactivate: deactivateFocusTrap,
-		} = useFocusTrap(kvLightbox);
+		} = useFocusTrap(trapElements, {
+			allowOutsideClick: true, // allow clicking outside the lightbox to close it
```
mcstover, line 229: Does this need to be conditional based on props sent to the lightbox? We don't allow outside clicks in certain circumstances such as employee verification for campaign pages.

emuvente: This won't bypass our existing `preventClose` logic. If preventClose is true, clicking on the background won't close the lightbox. This just allows clicking on the background to close the lightbox when preventClose is false.

mcstover (approved):

PR #323, opened 2023-12-01
feat: checkout transaction tracking

emuvente:
Replicating transaction tracking from UI for one-time checkouts.

PR #327, opened 2023-12-06
feat(oneTimeCheckout): redirect in promise so loading state remains active

emuvente:
Helps avoid an awkward experience of the checkout buttons appearing to complete moments before a new page loads.

### DIFF
--- a/@kiva/kv-shop/src/oneTimeCheckout.ts
+++ b/@kiva/kv-shop/src/oneTimeCheckout.ts
@@ -8,7 +8,8 @@ import { ShopError, parseShopError } from './shopError';
 import { callShopMutation, callShopQuery } from './shopQueries';
 import { validatePreCheckout } from './validatePreCheckout';
 import { wait } from './util/poll';
-import getVisitorID from './util/visitorId';
+import { getVisitorID } from './util/visitorId';
+import { redirectTo } from './util/redirect';
 import { getCheckoutTrackingData } from './receipt';
 
 interface CreditAmountNeededData {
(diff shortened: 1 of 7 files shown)


@kiva/kv-shop/src/oneTimeCheckout.ts
```diff
@@ -8,7 +8,8 @@ import { ShopError, parseShopError } from './shopError';
 import { callShopMutation, callShopQuery } from './shopQueries';
 import { validatePreCheckout } from './validatePreCheckout';
 import { wait } from './util/poll';
-import getVisitorID from './util/visitorId';
+import { getVisitorID } from './util/visitorId';
```
dyersituations, line 11: I can't comment directly on the line, but GH is saying `Unable to resolve path to module '@kiva/kv-analytics'`

mcstover (approved):

@kiva/kv-shop/src/oneTimeCheckout.ts
emuvente: Yeah, we've been having that error since [this PR](https://github.com/kiva/kv-ui-elements/pull/323#issuecomment-1836750401). We think maybe gha has a cached version of @kiva/kv-analytics? If you have any ideas what might be wrong there or how to fix it up, please share!

mcstover: Note there a separate import from kv-analytics that shows the same lint error for me locally... different method I think

PR #328, opened 2023-12-08
fix: p4 tag on loan card

eddieferrer:
in refactoring this and fixing some linting errors I created this bug


mcstover (approved):
Thank yoU!

PR #330, opened 2024-01-16
feat: voting category cards

michelleinez:


@kiva/kv-components/vue/KvVotingCard.vue
```diff
@@ -0,0 +1,147 @@
...
+				tw-text-h4
+				tw-inline-flex
+				tw-items-center"
+				style="padding: 2px 6px; text-transform: capitalize;"
+			>
+				<kv-material-icon
+					class="tw-h-2 tw-w-2"
+					:icon="mapMarkerIcon"
+				/>
+				{{ borrowerName }}, {{ country }}
```
mcstover, line 42: Do you think we'll need to truncate or remove the country in some cases do to being to long? I think there's a tw- class for that but don't recall exactly... It might be useful to create a story variation that tests what this looks like when the combination of the name and country spill out of the card width...

mcstover (approved):
Nice!

PR #334, opened 2024-01-24
feat: add new flexibleFundraisingEnabled filter

dyersituations:
https://kiva.atlassian.net/browse/ACK-867

- Added new filter from `ui` repo to shared filters
- Not currently used but ready for the future


roger-in-kiva (approved):
LG! Is there a default option? What is the current instead?

emuvente (approved):
Nice!

dyersituations:
> LG! Is there a default option? What is the current instead?

Default is `null`, like other filters. These types of filters also get the "all loans" by default in the implementation of the UI.

PR #339, opened 2024-02-01
feat: activity feed util updates

dyersituations:
https://kiva.atlassian.net/browse/ACK-924

- Updated `ActivityFeedService` methods while testing locally in CPS
- Focused on getting an activity and adding a comment
- Removed `KvCommentsContainer` since it ended up just being an empty wrapper of other shared components


mcstover (approved):
Seems good, Maybe it's ok to keep the container and it's tests for now if it reduces conflicts elsewhere...

dyersituations:
> Seems good, Maybe it's ok to keep the container and it's tests for now if it reduces conflicts elsewhere...

Yeah I went back and forth about removing the container. I added it back and tracked down why I was getting double events emitted (I guess the top element of the template gets the watchers as well, so we were double watching).

PR #341, opened 2024-02-13
fix: checkoutStatus query was only called once, the apollo cache retu…

mcstover:
…rns same response until timeout.

I was able to link this to Ui and test with some specialized steps b/c UI will not parse the dist or ts files directly:
1. Do the link setup
2. add the `--target es5` flag to the package.json build command
3. rebuild the repo
4. import methods into UI directly from each file `import { pollForFinishedCheckout } from '~/@kiva/kv-shop/dist/checkoutStatus';`

### DIFF
--- a/@kiva/kv-shop/src/checkoutStatus.ts
+++ b/@kiva/kv-shop/src/checkoutStatus.ts
@@ -39,6 +39,7 @@ export async function getCheckoutStatus(
 			transactionId: transactionSagaId,
 			visitorId: getVisitorID(),
 		},
+		fetchPolicy: 'network-only',
 	});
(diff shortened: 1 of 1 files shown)


dyersituations (approved):
Thanks for the local dev tips. In the few mins I had last week, I ran into normal npm link definitely not working.

emuvente (approved):

mcstover:
Thanks guys, this covers the immediate need and is probably ok for this particulary query. We could also set things up to pass through an `apolloOptions` object that we could include a `fetchPolicy` within among other options and spread onto the operation...

PR #342, opened 2024-02-16
feat: add invitationUrl to inviter param

christian14b:
Adding invitationUrl to the validatePreCheckout mutation inviter param.


roger-in-kiva:
I think the mutation should be updated to include the inviter here: https://github.com/kiva/kv-ui-elements/blob/1163285a4565cb678bd25f9483433866b4a08712/%40kiva/kv-shop/src/validatePreCheckout.ts#L9

Also, what is the `invitationUrl` param? The return from `executeOneTimeCheckout` method or the related to iwd lp?

mcstover (approved):

roger-in-kiva (approved):
👍

PR #343, opened 2024-02-22
fix: inviter param wasn't being used

dyersituations:


@kiva/kv-shop/src/validatePreCheckout.ts
```diff
@@ -6,15 +6,16 @@ import { ShopError, parseShopError } from './shopError';
 import type { ValetInviter } from './oneTimeCheckout';
 
 export const validatePreCheckoutMutation = gql`
-	mutation validatePreCheckout($basketId: String,
+	mutation validatePreCheckout(
+		$basketId: String,
 		$email: String,
 		$visitorId: String,
 		$emailOptIn: Boolean,
 		$inviter: InviterInput
```
mcstover, line 14: Here too?

dyersituations: It's already there. It just wasn't being used.

mcstover: oh right, sorry, thought this was a removal but it's an addition!

mcstover (approved):

PR #345, opened 2024-02-22
feat: forcing thanks page redirect in onetimecheckout method

roger-in-kiva:
- forcing thanks page redirect in `executeOneTimeCheckout` method


mcstover:
Why would be skipping transaction event tracking...

roger-in-kiva:
> Why would be skipping transaction event tracking...

Nice call! More than event tracking for logged in user is redirecting to portfolio, so maybe this fix should happen in `postPurchase` method right?

roger-in-kiva:
Closing this, thanks!

PR #348, opened 2024-02-28
feat: add activity feed update user method

dyersituations:
https://kiva.atlassian.net/browse/ACK-1023

- Apparently `getOrCreate` doesn't update if data is provided
- We need this `updateUser` method to be able to update existing users


roger-in-kiva (approved):
Nice!

mcstover (approved):

PR #349, opened 2024-02-28
feat: adding like to comment and remove reaction methods added

roger-in-kiva:
- Adding `addLikeToComment` and `removeReaction` methods to kv-activity-feed.

### DIFF
--- a/@kiva/kv-activity-feed/src/index.ts
+++ b/@kiva/kv-activity-feed/src/index.ts
@@ -108,4 +108,42 @@ export default class ActivityFeedService {
 
 		return success;
 	}
+
+	/**
+	 * Adds a like comment for the current user for the provided activity ID
+	 *
+	 * @param commentId The comment ID to add a like
+	 * @returns Whether the like was added successfully
+	 */
+	async addLikeToComment(commentId: string): Promise<boolean> {
+		let success = false;
+
+		try {
+			await this.client?.reactions.addChild('like', commentId);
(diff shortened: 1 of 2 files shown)


@kiva/kv-activity-feed/src/index.ts
```diff
@@ -108,4 +108,42 @@ export default class ActivityFeedService {
...
 		return success;
 	}
+
+	/**
+	 * Adds a like comment for the current user for the provided activity ID
+	 *
+	 * @param commentId The comment ID to add a like
+	 * @returns Whether the like was added successfully
+	 */
+	async addLikeToComment(commentId: string): Promise<boolean> {
```
dyersituations, line 118: Please add unit tests for these new methods, following the pattern of the other methods in this file.

roger-in-kiva: For this we need to update the `addComment` to return a valid commentId and add a child reaction then.
Said that, we also need to update the current flow we already have in CPS. Makes sense?

dyersituations: The unit test just needs to verify that we are using the correct 3rd party method, a real quick addition, like my most recent PR.

roger-in-kiva: Gotcha, thanks for checking.

dyersituations (approved):

PR #352, opened 2024-03-01
fix: remove const var import

roger-in-kiva:


@kiva/kv-components/vue/KvCommentsListItem.vue
```diff
@@ -86,10 +86,10 @@ import {
 import KvCommentsReplyButton from './KvCommentsReplyButton.vue';
 import KvCommentsHeartButton from './KvCommentsHeartButton.vue';
 import KvCommentsAdd from './KvCommentsAdd.vue';
-import { ADD_REACTION_EVENT } from './KvCommentsContainer.vue';
 
 export const REPLY_COMMENT_EVENT = 'reply-comment';
```
dyersituations, line 90: Will these exports cause the same problem?

roger-in-kiva: They are used in tests specs, so not related with other vue components...

dyersituations (approved):

PR #353, opened 2024-03-01
feat: add generic user avatar component

dyersituations:
- This component will be used in the new commenting components
- Ported from CPS and adjusted slightly

image
image
image


roger-in-kiva (approved):
Nice having this as shared component!

mcstover (approved):

PR #364, opened 2024-03-05
feat: saved searches feature not working

dyersituations:
https://kiva.atlassian.net/browse/CORE-1717

- The new filter is only supported in FLSS, so don't add to saved search

### DIFF
--- a/@kiva/kv-loan-filters/src/flexibleFundraisingEnabled.ts
+++ b/@kiva/kv-loan-filters/src/flexibleFundraisingEnabled.ts
@@ -77,11 +77,7 @@ export default {
 		return [];
 	},
 	getRemovedFacet: () => ({ flexibleFundraisingEnabled: null }),
-	getSavedSearch: (loanSearchState) => ({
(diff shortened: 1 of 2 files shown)


mcstover (approved):
In general, I'm hoping we can move toward Saved Searches linking to lend/filter and supporting all FLSS params but I realize there may be an interim state where things get messy.

roger-in-kiva (approved):

dyersituations:
> In general, I'm hoping we can move toward Saved Searches linking to lend/filter and supporting all FLSS params but I realize there may be an interim state where things get messy.

Agreed, the problem at the moment is that saving is in the monolith, and so we need to handle all filters explicitly there.

dyersituations:
> > In general, I'm hoping we can move toward Saved Searches linking to lend/filter and supporting all FLSS params but I realize there may be an interim state where things get messy.
> 
> Agreed, the problem at the moment is that saving is in the monolith, and so we need to handle all filters explicitly there.

A "better" solution will be to update the monolith endpoint to save this new filter and redirect to `/lend/filter`, but there are some legacy filters not supported in FLSS... 😿

PR #367, opened 2024-03-08
feat: inline activity feed components added

roger-in-kiva:
Screenshot 2024-03-08 at 12 16 04

Screenshot 2024-03-08 at 12 16 42

### DIFF
--- a/@kiva/kv-components/vue/KvInlineActivityCard.vue
+++ b/@kiva/kv-components/vue/KvInlineActivityCard.vue
@@ -1,7 +1,6 @@
 <template>
 	<div
-		class="tw-flex tw-items-center tw-gap-1 tw-py-0.5 tw-px-1 tw-shadow-lg"
-		style="border-radius: 8px;"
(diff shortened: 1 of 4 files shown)


@kiva/kv-components/vue/KvInlineActivityCard.vue
```diff
@@ -0,0 +1,61 @@
+<template>
+	<div
+		class="tw-flex tw-items-center tw-gap-1 tw-py-0.5 tw-px-1 tw-shadow-lg"
+		style="border-radius: 8px;"
```
dyersituations, line 4: Is there a tailwind class that's close enough?

roger-in-kiva: Or change to 4px, could make more sense for that card.

dyersituations: If 4px seems to make sense, go for it. We can generally use our existing styles in Tailwind.

roger-in-kiva: Very biased sometimes...

dyersituations (approved):

PR #368, opened 2024-03-12
fix: fix data reference for comment list item

dyersituations:
https://kiva.atlassian.net/browse/ACK-1076


roger-in-kiva (approved):
Thanks!

PR #372, opened 2024-03-19
fix: revert level 3 reactions and more specific error src

dyersituations:
- Turns out stream simply doesn't support reactions at this level 🤦 
- I sometimes see errors locally about the error import, so I made the import more specific


roger-in-kiva (approved):
Oh yeah, max nested level is 3...

dyersituations:
> Oh yeah, max nested level is 3...

I thought it would be configurable, but nope!

PR #374, opened 2024-03-21
fix: adjust commenting initially for chat

dyersituations:
- Starting to adjust components for new data from chat product (rather than activity feed product)
- This is just the first step, so I can start seeing the data format returned for replies
- More to come soon (stories currently broken)

### DIFF
--- a/@kiva/kv-components/tests/fixtures/mockFeedActivityData.js
+++ b/@kiva/kv-components/tests/fixtures/mockFeedActivityData.js
@@ -1,212 +1,155 @@
-const comments = {
-	duration: '6.42ms',
-	results: [
-		{
-			actor: {
-				created_at: '2024-02-01T19:48:26.890734Z',
-				updated_at: '2024-02-01T19:48:26.890734Z',
-				id: '123',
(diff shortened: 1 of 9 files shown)


@kiva/kv-components/tests/fixtures/mockFeedActivityData.js
```diff
@@ -1,212 +1,155 @@
...
-			},
-			object: 'test-challenge-id',
-			origin: null,
-			target: '',
-			time: '2024-02-01T17:54:24.932762',
-			verb: 'CHALLENGE',
+			],
+		},
+		children_counts: {
+			comment: 1,
```
roger-in-kiva, line 112: Do we have more than 1 comment here?

dyersituations: We can largely ignore this data for now, I just needed to get the tests to pass. There just isn't a great way to test these components locally with the live service. More changes coming when I know the data better!

roger-in-kiva (approved):

PR #396, opened 2024-05-01
feat: newly created global chip component [CIT-1429]

tiarascoleman:


@kiva/kv-components/vue/KvChip.vue
```diff
@@ -0,0 +1,73 @@
...
+			this.$emit('click-chip');
+		},
+	},
+};
+</script>
+
+<style lang="scss" scoped>
+
+	.chip {
+		border-radius: rem-calc(4);
```
dyersituations, line 50: How many of these styles can be replaced with tailwind equivalents?

tiarascoleman: @dyersituations  Likely will change convert line 64- I went through already and converted most. Unsure on the rest since the Tailwinds primitive classes caps at 4 rem

mcstover: My suggestion here is to use the "closest" available value from tailwind. It may look slightly different than the design but you can just make sure the label is balanced and looks good within the chip using our global sizes and it will be good.

@kiva/kv-components/vue/stories/KvChip.stories.js
```diff
@@ -0,0 +1,22 @@
...
+			<kv-chip
+			:title="chip-title"
+			/>
+		`,
+	});
+	template.args = args;
+	return template;
+};
+
+export const Default = story({ title: 'Chip Title' });
```
mcstover, line 22: Please add at least 1 more option with a longer title and with an email domain as well as a full email address.

tiarascoleman:
Currently what it looks like in Storybook: 
Screenshot 2024-05-02 at 11 08 22 AM
Screenshot 2024-05-02 at 11 08 34 AM
Screenshot 2024-05-02 at 11 08 48 AM

eddieferrer:
> Currently what it looks like in Storybook: Screenshot 2024-05-02 at 11 08 22 AM Screenshot 2024-05-02 at 11 08 34 AM Screenshot 2024-05-02 at 11 08 48 AM



> Currently what it looks like in Storybook: Screenshot 2024-05-02 at 11 08 22 AM Screenshot 2024-05-02 at 11 08 34 AM Screenshot 2024-05-02 at 11 08 48 AM

I think the X and the text need to be aligned better

@kiva/kv-components/vue/KvChip.vue
```diff
@@ -0,0 +1,79 @@
...
+		email: {
+			type: Boolean,
+			default: false,
+		},
+	},
+	emits: [
+		'click-chip',
+	],
+	setup() {
+		const closeButton = computed(() => mdiClose);
```
eddieferrer, line 68: This computed is unnecessary. You can just return `mdiClose` directly below and use it in the template. See: `@kiva/kv-components/vue/KvAccordionItem.vue` for an example

@kiva/kv-components/vue/KvChip.vue
```diff
@@ -0,0 +1,79 @@
...
+					{{ title }}
+				</div>
+				<div
+					v-else
+					class="tw-text-h4"
+				>
+					{{ title }}
+				</div>
+			</div>
+			<div class="tw-ml-2 tw-pt-1 tw-text-right">
```
eddieferrer, line 36: `tw-text-right` is unnecessary class

@kiva/kv-components/vue/KvChip.vue
```diff
@@ -0,0 +1,79 @@
...
+			tw-items-center
+			tw-flex
+			tw-flex-row
+			tw-cursor-pointer
+			tw-whitespace-nowrap
+			tw-rounded
+			tw-h-2
+			tw-py-2
+			tw-px-2
+			tw-mr-2
```
eddieferrer, line 16: Maybe we should not include these margin classes. This will allow the component that implements this component to handle that. For example, it would be easier to throw multiple chips in a flex layout if they don't have margin

@kiva/kv-components/vue/KvChip.vue
```diff
@@ -0,0 +1,79 @@
...
+					v-else
+					class="tw-text-h4"
+				>
+					{{ title }}
+				</div>
+			</div>
+			<div class="tw-ml-2 tw-pt-1 tw-text-right">
+				<kv-material-icon
+					:icon="closeButton"
+					class="tw-w-2 tw-h-2 tw-fill-gray-600"
```
eddieferrer, line 39: I don't believe this fill-gray class is doing anything

@kiva/kv-components/vue/KvChip.vue
```diff
@@ -0,0 +1,79 @@
...
+			tw-px-2
+			tw-mr-2
+			tw-mb-2
+			tw-select-none
+			"
+			@click="handleClick"
+		>
+			<div>
+				<div
+					v-if="!email"
```
eddieferrer, line 24: Thinking of how this component might be used in other contexts I don't think we should name this `email` -- I think a better approach might be to use a default slot. This would allow you to do things like: 
```<KvChip><h4>@email</h4></KvChip>``` or 
```<KvChip>Long Title</KvChip>```
And greatly simplify your component

File: @kiva/kv-components/vue/KvChip.vue [hunk 8]
```diff
@@ -0,0 +1,79 @@
+<template>
+	<div class="tw-inline-flex">
```
eddieferrer, line 2: To fix the alignment of the X and cleanup your dom I would remove your extraneous div's. I think the layout should be: 
```
<button>
<div></div>
<kv-material-icon/>
</button>
```

@kiva/kv-components/vue/KvChip.vue
tiarascoleman: @eddieferrer Are you referring to only the margin classes or the padding classes as well?

eddieferrer: the margin classes

tiarascoleman:
CIT-1429

Aligned!

eddieferrer (approved):

PR #404, opened 2024-05-20
feat: loan card primary and secondary button customization options

michelleinez:


michelleinez:
This only comprises part of the requirements for the changes to loan cards required by the employee engagement epic.
Contributions component will be part of a second PR.

@kiva/kv-components/vue/stories/KvClassicLoanCard.stories.js
```diff
@@ -397,3 +400,32 @@ export const ClickableTags = story({
...
+	photoPath,
+});
+
+export const RemoveButton = story({
+	loanId: loan.id,
+	loan,
+	kvTrackFunction,
+	photoPath,
+	basketItems: [{ id: loan.id, __typename: 'LoanReservation' }],
+	showRemoveButton: true,
```
mcstover, line 429: I don't think this prop is implemented... I htink you need to pass in the `secondaryButtonHandler` to activate this feature

@kiva/kv-components/vue/KvLendCta.vue
```diff
@@ -261,17 +273,20 @@ export default {
...
 				case 'loading':
 					return 'Loading...';
 				case 'refunded':
 				case 'expired':
 				default:
-					return 'Lend';
+					return this.primaryButtonText || 'Lend';
 			}
 		},
+		cta2ButtonText() {
```
dyersituations, line 287: It might be clearer to rename this and the `secondaryButton` usage to be specifically `checkoutButton`, since that's what we are using that CTA for.

@kiva/kv-components/vue/KvLendCta.vue
```diff
@@ -206,6 +206,18 @@ export default {
...
 		},
+		primaryButtonText: {
+			type: String,
+			default: 'Lend',
+		},
+		secondaryButtonText: {
+			type: String,
+			default: 'Checkout now',
+		},
+		secondaryButtonHandler: {
```
dyersituations, line 217: What is the use case for this "alternative checkout functionality"? Could we use an emit here instead?

michelleinez: We use this to emit an event that removes the loan from the basket.

michelleinez: https://www.figma.com/design/fvwjgrVzEEKCskC4Cn2cI3/Impact-Dashboard-page-v2---Kiva%40Work---Q124---SH?node-id=924-4785&t=8hgxzot9dgCvOcMW-0

mcstover: My first impression was to also use an emit here but I also see that we're passing in a function for several other features so we've kind of mixed the approaches into this component... Either way we have to "handle" the operation in the parent...so I think it's fine for now.

@michelleinez Please keep this in mind once you get to integrating with the [LoanCardContainer](https://github.com/kiva/cms-page-server/blob/main/components/loan/cards/KvClassicLoanCardContainer.vue) in CPS...it may be that it's easier to manage the email approach...

michelleinez: Using an emit here instead of passing in the function wouldn't be able to override the existing functionality when the function is defined since the data is only one way from child to parent.  We need to alter child state as well if the function is defined to not go through the checkout process.

@kiva/kv-components/vue/KvLendCta.vue
mcstover: I agree this naming is getting a little wonky...generally ok with this for now but I wonder if we should break it down by "stages" at somepoint... ie. ctaButtonText > stageOneButtonText + cta2ButtonText > stageTwoButtonText or something along those lines... or maybe lendButtonText and loanReservedButtonText...

@kiva/kv-components/vue/stories/KvClassicLoanCard.stories.js
michelleinez: done

@kiva/kv-components/vue/KvLendCta.vue
michelleinez: What about loanInBasketButtonText?

michelleinez: ok done

@kiva/kv-components/vue/KvLendCta.vue
mcstover: It's true we'd need to also conditionally stop the checkout action. We'd need to refactor the `externalLinks` prop usage to stop the navigation...

Looking at the current button usage which has a `@click.native` handler we may also need to include `preventDefault()` in the handler function we pass in to keep the click from going out to checkout...

mcstover (approved):

mcstover (approved):

dyersituations (approved):

PR #407, opened 2024-05-29
feat: remove secondary button default behavior when custom behavior is set

michelleinez:


mcstover (approved):
Were you able to use preventDefault in your passed in handler to prevent the click/transition?

michelleinez:
> Were you able to use preventDefault in your passed in handler to prevent the click/transition?

I wasn't...  I tried putting it there but the link to basket was still coming up.

PR #408, opened 2024-05-30
feat: return shop data for applyPromoCredit including errors

michelleinez:


@kiva/kv-shop/src/basketCredits.ts
```diff
@@ -54,9 +54,9 @@ export interface ApplyPromoCreditData {
 export async function applyPromoCredit(
 	apollo: ApolloClient<any>,
 	options: MutationOptions<any>,
-): Promise<boolean> {
-	if (!options?.variables?.creditType && !options?.variables?.redemptionCode) {
-		return Promise.resolve(false);
+): Promise<any> {
+	if (!options?.variables?.creditType || !options?.variables?.redemptionCode) {
```
mcstover, line 58: This is not an either or scenario. Both must be present or the mutation will fail.

@kiva/kv-shop/src/basketCredits.ts
```diff
@@ -74,7 +74,7 @@ export async function applyPromoCredit(
 		variables: { ...options?.variables },
 	});
 
-	return !!data?.shop?.addCreditByType;
+	return data?.shop;
```
mcstover, line 77: This is going to need to return `data` at the root b/c the `errors` object will live alongside the `shop` node...
ex.
```
{"data":{"shop":{"id":"GQiC3v8YCSq73smPLsQLXQ==","addCreditByType":null,"__typename":"ShopMutation"}},"errors":[{"message":"The promotion for (dogs) does not exist, but you can still use your own credit to checkout.","locations":[{"line":1,"column":146}],"path":["shop","addCreditByType"],"extensions":{"code":"upc.promoDoesNotExist"}}]}
```

@kiva/kv-shop/src/basketCredits.ts
michelleinez: Don't we want the function to return false whether the credit type OR the redemption code is missing?
This is a negation so || seems like the right operator, not &&.  && would be if we weren't negating.

That is, if both must be present for the mutation not to fail...

@kiva/kv-shop/src/basketCredits.ts
michelleinez: done

mcstover: Oh shoot... we actually have to return the full response b/c `data` and `errors` are siblings :(

@kiva/kv-shop/src/basketCredits.ts
```diff
@@ -54,9 +54,9 @@ export interface ApplyPromoCreditData {
 export async function applyPromoCredit(
 	apollo: ApolloClient<any>,
 	options: MutationOptions<any>,
-): Promise<boolean> {
-	if (!options?.variables?.creditType && !options?.variables?.redemptionCode) {
-		return Promise.resolve(false);
+): Promise<any> {
+	if (!options?.variables?.creditType || !options?.variables?.redemptionCode) {
+		return false;
```
mcstover: Still need to return the promise here... probably should return a shop error structure. Something like...
```
{
  "errors":
    [
      {
        "message":"Missing require parameter.",
        "extensions":{"code":"upc.missing_parameter"}
      }
    ]
}
```

@kiva/kv-shop/src/basketCredits.ts
michelleinez: Otherwise we'd be requiring that both be missing in order for the mutation to return false.

mcstover: Yea I guess it will work either way. If either of these are missing we'll return...

@kiva/kv-shop/src/basketCredits.ts
michelleinez: done

@kiva/kv-shop/src/basketCredits.ts
mcstover: Good catch on this!

@kiva/kv-shop/src/basketCredits.ts
mcstover: I got that response by using a fake upc code to force failure in our existing test page. you can inspect the network panel to find that call and it's result.

@kiva/kv-shop/src/basketCredits.ts
```diff
@@ -74,7 +82,7 @@ export async function applyPromoCredit(
 		variables: { ...options?.variables },
 	});
 
-	return !!data?.shop?.addCreditByType;
+	return Promise.resolve(data);
```
mcstover, line 85: I don't know that we need to wrap this one b/c `data` is the result of the query promise...

michelleinez: done

mcstover (approved):

PR #412, opened 2024-06-04
feat: allow return of full error message in kv-shop for applyPromoCredit

michelleinez:


@kiva/kv-shop/src/shopError.ts
```diff
@@ -91,6 +91,23 @@ export function parseShopError(error: any) {
 		}, errorMessage);
 	}
 
+	if (
+		errorCode === 'upc.promoNotYetStarted'
+		|| errorCode === 'upc.promoHasEnded'
```
mcstover, line 96: please only look for the inclusion of `upc` within the `errorCode` and not the rest of the specific 2nd segment of code... There are several others and I don't think we should be copying those codes here to check against...

mcstover: oh I see you have it below, so you can just drop this first block...

@kiva/kv-shop/src/shopError.ts
```diff
@@ -91,6 +91,23 @@ export function parseShopError(error: any) {
...
+			original: error,
+		}, errorCode);
+	}
+
+	if (errorCode.startsWith('upc')) {
+		return new ShopError({
+			code: errorCode,
+			original: error,
+		}, errorMessage);
+	}
```
mcstover, line 99: This block should cover all errorCode(s) prefixed with `upc`.

mcstover (approved):

@kiva/kv-shop/src/shopError.ts
mcstover: Also noting we'll have to make another broad exclusion for actual promo codes sometime in the future when we want those working on the impact dashboard too.

michelleinez: @matthews is the new version I just pushed okay?  Because what you’re suggesting in git is not going to get me the error code, so how would I use the error to gate pre & post campaign state?

mcstover: Please add a comment above to state that this allows us to get the full error message for all `upc.` failures and notices.

michelleinez: Currently what the errorMessage returns is ShopError: An unknown error occurred.
But I think errorCode should have the part that says upc.promoHasEnded and upc.promoNotYetStarted

mcstover: Also note, we will not want to use this `${errorCode}: ${errorMessage}` in our tip message / presentation. We'll just use the `message` value which plainly states the problem.

mcstover: So, we'll probably want to change that last value to just `${errorMessage}` instead of a string with both included.

michelleinez: But then how could we use the error to gate pre and post campaign state if the errorCode is missing?  @mcstover

mcstover: You can pull the nested code and error out using the object keys:
`console.log('upcResult', upcResult, typeof upcResult, upcResult.code, upcResult.message);`
results in:
```
upcResult
ShopError: An unknown error occurred.
...stacktrace
object
shop.unknown
An unknown error occurred.
```

mcstover: `upcResult.original`
```
{
  "message": "The promotion for (hotdogs) does not exist, but you can still use your own credit to checkout.",
  "locations": [
    {
      "line": 1,
      "column": 146
    }
  ],
  "path": [
    "shop",
    "addCreditByType"
  ],
  "extensions": {
    "code": "upc.promoDoesNotExist"
  }
}
```

mcstover (changes requested):
I think this can be closed, we have what we need by inspecting the `response.original` field.

PR #414, opened 2024-06-06
fix: ensure pass distributionModel value is capitalized when lowercas…

mcstover:
…e, converted from legacy

Distribution model in legacy is passed as camel case. This ensures that when we start with a loan search criteria and create loanSearchState with it, we automatically move to uppercase for this enum.

I didn't see any other filters that have this issue...

### DIFF
--- a/@kiva/kv-loan-filters/src/tests/distributionModels.spec.ts
+++ b/@kiva/kv-loan-filters/src/tests/distributionModels.spec.ts
@@ -115,6 +115,11 @@ describe('distributionModels.ts', () => {
 				expect(distributionModels.getFlssFilter({ distributionModel: 'DIRECT' }))
 					.toEqual({ distributionModel: { eq: 'DIRECT' } });
 			});
+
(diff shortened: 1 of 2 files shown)


@kiva/kv-loan-filters/src/tests/distributionModels.spec.ts
```diff
@@ -115,6 +115,11 @@ describe('distributionModels.ts', () => {
 				expect(distributionModels.getFlssFilter({ distributionModel: 'DIRECT' }))
 					.toEqual({ distributionModel: { eq: 'DIRECT' } });
 			});
+
+			it('should capitalize filters', () => {
+				expect(distributionModels.getFlssFilter({ distributionModel: 'direct' }))
+					.toEqual({ distributionModel: { eq: 'DIRECT' } });
+			});
```
mcstover, line 122: This test failed with the current state and was resolved with the updates in the module.

dyersituations (approved):

PR #416, opened 2024-06-06
feat: implement custom loan details on loan card

eddieferrer:
We need a way to control what happens when you click on the card elements (the image, the loan-use, the progress meter). I wanted to avoid adding yet another prop. This makes it so that if the prop "custom-loan-details" is true, emit an event instead of navigating. 

I'm pretty confident this prop is not used anywhere, and was just copied over from: `src/components/LoanCards/KivaClassicBasicLoanCard.vue`, I searched and could not find a use of it. Also, the previous use of `this.$emit` was throwing an error and not working. this fixes that

### DIFF
--- a/@kiva/kv-components/vue/KvClassicLoanCard.vue
+++ b/@kiva/kv-components/vue/KvClassicLoanCard.vue
@@ -25,7 +25,7 @@
 						:href="readMorePath"
 						class="tw-flex"
 						aria-label="Borrower image"
-						@click="clickReadMore('Photo')"
+						@click.native="clickReadMore('Photo', $event)"
 					>
 						<kv-borrower-image
 							class="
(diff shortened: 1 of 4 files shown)


@kiva/kv-components/vue/KvClassicLoanCard.vue
```diff
@@ -93,7 +93,7 @@
 					class="tw-flex hover:tw-no-underline focus:tw-no-underline"
 					:class="{ 'tw-px-1': largeCard }"
 					aria-label="Loan tag"
-					@click="clickReadMore('Tag')"
+					@click="clickReadMore('Tag', $event)"
```
dyersituations, line 96: Does this need `click.native` like the others?

mcstover (approved):
Good find and fix!

dyersituations (approved):

PR #418, opened 2024-06-14
fix: use auth0 package file name to be compatible with es modules

dyersituations:
- Apparently this is a known issue with using the Apollo package within an ES module
- The Vue migration converted the `ui` repo to an ES module, so this changeset is needed (checkout page currently crashes the dev server 😄 )

### DIFF
--- a/@kiva/kv-shop/src/basket.ts
+++ b/@kiva/kv-shop/src/basket.ts
@@ -1,5 +1,4 @@
-import type { ApolloClient } from '@apollo/client/core';
-import { gql } from '@apollo/client/core';
+import { gql, type ApolloClient } from '@apollo/client/core/core.cjs';
 import { getCookieValue, setCookieValue } from './util/cookie';
 import { parseShopError } from './shopError';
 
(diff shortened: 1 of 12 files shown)


emuvente (approved):

mcstover:
Does this work with the current UI repo state?

This may also affect how we "configure" [kv-shop in CPS](https://github.com/kiva/cms-page-server/blob/main/nuxt.config.ts#L92)...

dyersituations:
> Does this work with the current UI repo state?
> 
> This may also affect how we "configure" [kv-shop in CPS](https://github.com/kiva/cms-page-server/blob/main/nuxt.config.ts#L92)...

I do wonder about this too, and I will check both repos this morning.

dyersituations:
@mcstover Between meetings and errands, tracked down a workaround to get kv-shop change here to work in the CPS repo: https://github.com/kiva/cms-page-server/pull/1305. These changes worked as is in the `ui` repo, both old, and in the `vue-3` branch.

PR #419, opened 2024-06-25
fix: max dimensions on kvContentfulImg

eddieferrer:
Contentful will not serve images with a single dimension > 4000 px. This came up again. Usually an issue on retina screens when we build the 2x url. I made a similar fix for the responsive image sets in cps, but we need this fix for embedded images.  This seems to be coming up lately because the imported images from the legacy blog are huge, and we are linking to blog entries in other contexts for example the new us-hubs page

### DIFF
--- a/@kiva/kv-components/vue/KvContentfulImg.vue
+++ b/@kiva/kv-components/vue/KvContentfulImg.vue
@@ -213,9 +213,22 @@ export default {
 
 		const buildUrl = (image = null, multiplier = 1) => {
 			let src = image && image.url ? `${image.url}?` : `${contentfulSrc.value}?`;
-			const imgWidth = image ? image.width : width.value;
-			const imgHeight = image ? image.height : height.value;
-
+			let imgWidth = image ? image.width : width.value;
+			let imgHeight = image ? image.height : height.value;
+			// The max contentful image size is 4000px so we have to
+			// impose a limit of 2000px here for both height and width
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/vue/KvContentfulImg.vue
```diff
@@ -213,17 +213,32 @@ export default {
...
+			let imgWidth = (image ? image.width : width.value) * multiplier;
+			let imgHeight = (image ? image.height : height.value) * multiplier;
+
+			// The max contentful image size is 4000px so we have to
+			// impose a limit of 2000px here for both height and width
+			// so when we request the retina x2 image we don't go over the 4000px limit
+			let newMultiplier;
+			if (imgWidth >= 2000) {
+				newMultiplier = imgWidth / 2000;
+				imgWidth = 1999;
```
dyersituations, line 225: Should `2000` and `1999` match here to ensure we aren't potentially off by 1px?

eddieferrer: Party like its 1999

dyersituations: Haha.

Did something change with the last force push? Or does it not matter as much that the multiplier doesn't use 1999?

eddieferrer: Sorry, fixed

mcstover (approved):

PR #422, opened 2024-07-09
feat: brown theme added to tokens

roger-in-kiva:
- brown theme added to tokens

### DIFF
--- a/@kiva/kv-tokens/configs/kivaColors.cjs
+++ b/@kiva/kv-tokens/configs/kivaColors.cjs
@@ -7,7 +7,7 @@ const {
 	'dark-green': darkGreenThemeTokens,
 	mint: mintThemeTokens,
 	'dark-mint': darkMintThemeTokens,
-	brown: brownThemeTokens,
+	'dark-stone': darkStoneThemeTokens,
 } = designtokens.colors.theme;
 
 /**
@@ -37,7 +37,7 @@ const darkTheme = buildCSSVarsFromTokens(darkThemeTokens);
 const mintTheme = buildCSSVarsFromTokens(mintThemeTokens);
 const darkGreenTheme = buildCSSVarsFromTokens(darkGreenThemeTokens);
 const darkMintTheme = buildCSSVarsFromTokens(darkMintThemeTokens);
-const brownTheme = buildCSSVarsFromTokens(brownThemeTokens);
+const darkStoneTheme = buildCSSVarsFromTokens(darkStoneThemeTokens);
 
 // function to allow background opacity and text opacity with tailwind colors
 // https://www.youtube.com/watch?v=MAtaT8BZEAo
@@ -89,5 +89,5 @@ module.exports = {
 	darkGreenTheme,
 	mintTheme,
 	darkMintTheme,
-	brownTheme,
+	darkStoneTheme,
 };
--- a/@kiva/kv-tokens/configs/kivaColors.cjs
+++ b/@kiva/kv-tokens/configs/kivaColors.cjs
@@ -7,6 +7,7 @@ const {
 	'dark-green': darkGreenThemeTokens,
 	mint: mintThemeTokens,
 	'dark-mint': darkMintThemeTokens,
+	brown: brownThemeTokens,
 } = designtokens.colors.theme;
 
 /**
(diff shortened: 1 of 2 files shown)


@kiva/kv-tokens/configs/kivaColors.cjs
```diff
@@ -7,6 +7,7 @@ const {
 	'dark-green': darkGreenThemeTokens,
 	mint: mintThemeTokens,
 	'dark-mint': darkMintThemeTokens,
+	brown: brownThemeTokens,
```
dyersituations, line 10: Is "brown" the name being used by design?

roger-in-kiva: Nice call, I'm not sure, `Stone` maybe?

https://www.figma.com/design/61wzqlnIHQyZezhkfGkXbO/Kiva-2.0-Working-File-Q2.2204.[SH]?node-id=523-1268&m=dev

dyersituations: @mcstover do you happen to know the theme name here?

mcstover: Not sure off hand but we should def get clarity from design on the name of the themes as they will eventually be showing up in Contentful so they should be "official"

roger-in-kiva: Ok, based on this slack convo https://kivahq.slack.com/archives/CB7GRAFKK/p1720567136303099?thread_ts=1720563250.654589&cid=CB7GRAFKK

Name for this theme is Dark Stone.

I agree not having an incomplete theme, but looks like there is some extra work left from design team for this.

We can go ahead with this temporary case to progress with the carousel work?

@kiva/kv-tokens/primitives.json
```diff
@@ -163,6 +163,15 @@
 				"background": {
 					"primary": "#454545"
 				}
+			},
+			"dark-stone": {
+				"text": {
+					"primary": "#F5F5F5",
+					"action": "#F5F5F5"
```
mcstover, line 170: Shouldn't this be `#F3F1EF` ? That is the lightest Stone theme color. In some of the ecosystem design examples the lightest color in the theme is used for the text color when over the dark color as a bg.

mcstover: https://www.figma.com/design/ns8HGt1yjQBfiC1OKlHU40/Legacy---%F0%9F%8C%BF-Ecosystem---Kiva-Brand-2022?node-id=3077-2580&t=ZAQmb3Azw5Bd20Mm-0

mcstover: We may also be able to use some values from the new "Default" dark theme that's in progress to guide some of the other variables. In particular, setting text-action to the same value means links would be indistinguishable...
https://www.figma.com/design/ns8HGt1yjQBfiC1OKlHU40/Legacy---%F0%9F%8C%BF-Ecosystem---Kiva-Brand-2022?node-id=4315-2579&t=ZAQmb3Azw5Bd20Mm-0

roger-in-kiva: That's a nice call but we have `#F5F5F5` in figma, said that, should this be a new theme different than Stone?
Or maybe other call to design to follow our current themes?

emuvente:
I think the brown is actually the marigold color set, not stone.

@kiva/kv-tokens/primitives.json
emuvente: the figma design I see uses stone-4 and marigold. I don't think it makes sense to name this theme based on either of those.

roger-in-kiva:
> I think the brown is actually the marigold color set, not stone.
> 
> 
> 
> edit: looks like the figma design has both stone and marigold, so this probably needs a different name than either of those if we're keeping it.

Thanks @emuvente. Dark Stone was suggested by design team, do you have a suggestion to name this based on your comment?

roger-in-kiva:
Compared with this new theme and the current Dark theme, only value we are changing is for brown background color.
Said that, could we use the current dark theme and override the bg color for the new carousel?

mcstover (approved):
I think this is fine for now... BG color will be customizable and your type styles will be the light color defined for text. We can refine them later...

Please import and setup a story for this new theme so we can look at it against the elements it affects. https://github.com/kiva/kv-ui-elements/blob/main/%40kiva/kv-components/vue/stories/KvThemeProvider.stories.js#L151

mcstover:
@roger-in-kiva the background color should come from setting up a Background in Contentful so it can be anything we'd like.

mcstover:
In general, we should not be hard-coding a background color into a shared reusable contentful component...

PR #425, opened 2024-07-12
fix: modal position and size on desktop

christian14b:
image

Design on desktop: (Missing check in dekstop)
image

Mobile keeps being full width and in the top of the page


dyersituations (approved):

emuvente (commented):
👍

PR #426, opened 2024-07-12
feat: dark stone theme provider added

roger-in-kiva:
- Dark Stone theme provider added

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -55113,7 +55113,7 @@
         "@babel/core": "^7.14.8",
         "@babel/eslint-parser": "^7.13.14",
         "@babel/preset-env": "^7.14.8",
-        "@kiva/kv-tokens": "*",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -55113,7 +55113,7 @@
         "@babel/core": "^7.14.8",
         "@babel/eslint-parser": "^7.13.14",
         "@babel/preset-env": "^7.14.8",
-        "@kiva/kv-tokens": "^2.12.0",
+        "@kiva/kv-tokens": "*",
```
dyersituations, line 55116: What caused this change? Should it be reverted?

roger-in-kiva: Oh, I didn't see that. Just updated the package as usual.
I guess should be reverted...

dyersituations (approved):

PR #430, opened 2024-07-18
feat: brush-stroke underline for headings mvp MP-348

emuvente:
Brush stroke underline style for headings. This first round is kiva-green only, with additional colors to be added later via themes.

My goals with this were:

- use an SVG as an underline
- clip/mask the SVG to round off the ends (optional)
- support underlining multi-line text
- change the color based on the theme
- minimize performance impact

Due to various CSS limitations, the only method I could find to support this is a [sprite-sheet style SVG](https://css-tricks.com/svg-fragment-identifiers-work/) background image and a [CSS variable defining the full url](https://blog.coderfy.io/css-variables-dont-work-with-the-url-function) for that image and the target color, e.g. `--heading-underline-primary: url('/heading-underline.svg#2AA967')`.

This technique requires hosting the SVG separately; it [can't use SVGs defined in the HTML document](https://www.svgbackgrounds.com/how-to-add-svgs-with-css-background-image/#:~:text=There%20are%20TWO%20methods). Since the initial ask is just for kiva-green, I have made a separate ticket for figuring out how/where to host the SVG, and I defined the css variable as a base64-encoded inline SVG with an explicit #2AA967 fill.

### Desktop
Screenshot 2024-07-18 at 5 06 25 PM

### Tablet
Screenshot 2024-07-18 at 5 06 45 PM

### Mobile
Screenshot 2024-07-18 at 5 07 10 PM


dyersituations (approved):

mcstover (approved):
Looks like a nice expandable approach that can flow forward with our theming work.

We may need to turn off underline styles in our cms-page-server code for dynamic hero and story cards to we don't conflict with those existing underline styles.

PR #435, opened 2024-07-26
feat: remove clickable tags exp

christian14b:


@kiva/kv-components/vue/KvIntroductionLoanCard.vue
```diff
@@ -228,6 +228,7 @@
 					:loan="loan"
 					variation="matched-loan"
 					class="tw-text-center !tw-text-brand"
+					:kv-track-function="kvTrackFunction"
```
dyersituations, line 231: Why was this added?

dyersituations: Please remove the prop from `KvLoanTag`, since you're right, it's no longer being used.

dyersituations (approved):

PR #436, opened 2024-08-02
chore: add loan-card class to all loan cards for cypress test access

mcstover:
Our cypress tests search through collections of loan cards to target specific states. This will make sure there is always a way to find a list of loan cards on a page across any usages of these components.


eddieferrer:
Would `data-testid` be more appropriate?

mcstover:
[Ok I verified that we can query/get/find lists using data-testid too](https://testing-library.com/docs/queries/about)

eddieferrer (approved):

PR #437, opened 2024-08-05
fix: relative node_modules path broke when imported into another app

dyersituations:
https://kiva.atlassian.net/browse/MP-607

- The relative path to `node_modules` didn't work when imported into a different repo
- I went with the alias we are already using in CPS


mcstover (approved):
Thanks!

eddieferrer (approved):

PR #441, opened 2024-08-14
feat: country layers added to leaflet map

roger-in-kiva:
- country layers added to leaflet map

Screenshot 2024-08-14 at 11 37 29 a m
Screenshot 2024-08-14 at 11 37 42 a m
Screenshot 2024-08-14 at 11 37 57 a m


@kiva/kv-components/data/countries-borders.json
dyersituations, line 1: Just out of curiosity, how did you solve the file-size problem? It looks like this file is 477KB. The file in the `ui` repo appears to be 23MB. This is a awesome improvement!

roger-in-kiva: @mcstover found this 😎; is a low resolution, but working pretty similar as monolith case

mcstover: We did a lot of searching and discovery of different precompiled country polygon data sources and found a couple that were lower fidelity. The borders aren't pixel perfect against a real map and there is less meta data such as additional languages for names and such. We also found a pretty strong higher fidelity version that comes in around 4mb.

With the primary task here being to replicate the existing page which is also "low res", this seemed like a fine compromise for now.

@kiva/kv-components/tests/unit/specs/utils/mapUtils.spec.js
```diff
@@ -1,6 +1,7 @@
 import {
 	getCoordinatesBetween,
-} from '../../../../utils/mapAnimation';
+	getLoansIntervals,
+} from '../../../../utils/mapUtils';
 
 describe('mapAnimation', () => {
```
dyersituations, line 6: This block should be renamed to `mapUtils`, and the new describe below removed, the `describe('getLoansIntervals', () => {` moved inside this `describe`.

@kiva/kv-components/utils/mapUtils.js
```diff
@@ -270,3 +270,41 @@ export function animationCoordinator(mapInstance, borrowerPoints) {
 		flyToPoint(currentPointIndex);
 	});
 }
+
+/**
+ * This function returns an array of not overlapped intervals between min and max
+ * @param {Integer} min - min number of the interval
+ * @param {Integer} max - max number of the interval
+ * * @param {Integer} nbIntervals - number of intervals
```
dyersituations, line 278: Extra asterisk

@kiva/kv-components/vue/KvMap.vue
```diff
@@ -408,6 +505,96 @@ export default {
...
+			const layer = e.target;
+			const { feature } = layer;
+
+			layer.setStyle({
+				fillColor: this.getCountryColor(feature.lenderLoans),
+			});
+		},
+		circleMapClicked(countryIso) {
+			this.$emit('country-lend-filter', countryIso);
+			this.$router.push({
```
dyersituations, line 591: Given that the router isn't defined in the component library, this click method should probably just emit.

@kiva/kv-components/vue/KvMap.vue
```diff
@@ -408,6 +505,96 @@ export default {
...
+		},
+		countryStyle(feature) {
+			return {
+				color: kvTokensPrimitives.colors.white,
+				fillColor: this.getCountryColor(feature.lenderLoans),
+				weight: 1,
+				fillOpacity: 1,
+			};
+		},
+		getCountryColor(lenderLoans) {
```
dyersituations, line 529: Could this method use some unit tests as well? I know there's some randomness to the colors, so I understand if a unit test might not be as reliable.

@kiva/kv-components/data/countries-borders.json
dyersituations: Thanks for the info @mcstover

@kiva/kv-components/vue/KvMap.vue
roger-in-kiva: I added one for default color.

dyersituations (approved):

mcstover (commented):
Nice work!

PR #444, opened 2024-08-16
feat: dotted version for carousel controls

roger-in-kiva:
- dotted version for carousel controls

Screenshot 2024-08-16 at 1 31 52 p m
Screenshot 2024-08-16 at 1 32 04 p m

### DIFF
--- a/@kiva/kv-components/vue/KvCarousel.vue
+++ b/@kiva/kv-components/vue/KvCarousel.vue
@@ -75,6 +75,26 @@
 				<span class="tw-sr-only">Show next slide</span>
 			</button>
 		</div>
+		<!-- Dotted Controls -->
+		<div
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/vue/KvCarousel.vue
```diff
@@ -119,6 +119,26 @@
 				</button>
 			</div>
 		</template>
+		<!-- Dotted Controls -->
+		<div
+			v-if="slideIndicatorCount > 1 && isDotted"
```
dyersituations, line 124: This should probably be `v-else-if`

roger-in-kiva: Makes sense!

dyersituations (approved):

PR #445, opened 2024-08-19
fix: vite has some complications around importing json

dyersituations:
- Working on the Vue migration encountered an issue with Vite and importing json in the util file
- I tested this locally in the `node_modules` folder manually and it seems to work fine


mcstover (approved):
This update is fine. We should also be able to use a slightly more complex import to ensure this could work...though there may be some other "linkage" we need to provide for the separate dependency...

ex. `import * as kvTokensPrimitives from '@kiva/kv-tokens/primitives.json?raw';`

PR #458, opened 2024-09-06
fix: handle slide scale instead of img

christian14b:
Scaling the whole slide instead of the container and prevent overflow hidden in the main slide due to over scale


dyersituations:
@christian14b could you post a screenshot of before and after?

dyersituations (approved):

PR #459, opened 2024-09-13
fix: remove computed from next tick

dyersituations:
https://kiva.atlassian.net/browse/MP-780

- Hoping this fixes the memory leak


eddieferrer (approved):
Certainly an improvement even if it doesnt.

mcstover (approved):

PR #464, opened 2024-10-04
fix: pie chart not animating in vue 3

emuvente:
The `activeSlice === slice` comparison in the template works in Vue 2, but in Vue 3 `activeSlice` is actually a Ref of the slice and so `activeSlice.value` needs to be compared instead. For compatibility with Vue 2, I moved the comparison out of the template and into a method.

I also change the check for the large slice to be for slices larger than 99% rather than exactly 100%, to account for any rounding of numbers less than 100 that might result in the slice not displaying.


roger-in-kiva (approved):
Great for the large slice! 🤌

PR #466, opened 2024-10-08
feat: new voting card

eddieferrer:
New voting card component with story

Screenshot 2024-10-08 at 1 11 46 PM
Screenshot 2024-10-08 at 1 11 29 PM

### DIFF
--- a/@kiva/kv-components/vue/KvVotingCardV2.vue
+++ b/@kiva/kv-components/vue/KvVotingCardV2.vue
@@ -1,0 +1,98 @@
+<template>
+	<div
+		class="vote_card"
+		:style="cssProps"
+	>
+		<h2 class="tw-italic tw-pb-1">
+			{{ title }}
+		</h2>
+		<p class="tw-pb-1.5">
+			{{ description }}
+		</p>
+		<div class="tw-block md:tw-flex tw-justify-between">
+			<div
+				v-if="showPercentage"
+				class="tw-flex tw-items-center tw-w-full tw-max-w-16"
+			>
+				<kv-progress-bar
+					class="tw-flex-grow"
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/vue/KvVotingCardV2.vue
```diff
@@ -0,0 +1,98 @@
...
+	computed: {
+		cssProps() {
+			return {
+				'--background-image': `url(${this.backgroundImageUrl})`,
+			};
+		},
+	},
+	methods: {
+		castVote() {
+			this.$emit('vote');
```
mcstover, line 83: Don't we need to return or declare the emits?

eddieferrer: I think this works with the options API, changed it to use setup(), added emit test in story.

mcstover (approved):

@kiva/kv-components/vue/KvVotingCardV2.vue
```diff
@@ -0,0 +1,111 @@
...
+		'vote',
+	],
+	setup(props, { emit }) {
+		const {
+			backgroundImageUrl,
+		} = toRefs(props);
+
+		const cssProps = computed(() => {
+			return {
+				'--background-image': `url(${backgroundImageUrl.value})`,
```
mcstover, line 88: Do we need to namespace this css variable so it's doesn't conflict with others on the page? Or will it be scoped into the component instance somehow?

eddieferrer: Maybe scoped? I'm not sure, no harm in namespacing it, thats a good idea

mcstover (approved):

PR #469, opened 2024-10-23
feat: embla carousel version updates

tiarascoleman:
This PR is meant to bring our version of embla up to at least V7 which supports their new plugin architecture. We're trying to move to their latest version, from 4.3.x > 8.3.x.


mcstover:
My current read on the test failures and some local testing is that the component isn't rendering for Vue3 so we'll need to do some additional work to fix things up.

PR #474, opened 2024-11-04
fix: centralize braintree helper text

dyersituations:
https://kiva.atlassian.net/browse/MP-994

- Moved braintree helper text to component library


emuvente (approved):
Thanks!

PR #475, opened 2024-11-04
feat: compile kv-component utils

eddieferrer:
* Add compiling to utils folder of kv-components to avoid issues when installed in other repos. 
* Also removes gap in vertical carousel to make it a little more versatile


eddieferrer:
@emuvente Does this look correct?

emuvente (approved):
Yeah, I think this should work

dyersituations (approved):

mcstover (approved):

PR #476, opened 2024-11-05
fix: add components to dist folder as well

eddieferrer:
This is an attempt to try to move the components into the dist folder and make this work. 

~That being said, when I copied this over into CPS node_modules, it doesnt seem to work.~

This now seems to be working, pending a vue-demi error, but I think actually npm installing this into cps might fix that.


dyersituations (approved):

mcstover (approved):

eddieferrer:
@emuvente still doesnt work :(

eddieferrer:
> @emuvente still doesnt work :(

Nevermind I got it working in: https://github.com/kiva/kv-ui-elements/pull/477

PR #484, opened 2024-12-02
fix: removing margin and fixed position

christian14b:
Removing specific margin and fixed position from sheet

image


roger-in-kiva:
Should we consider a small margin at top as current my kiva menu?
Screenshot 2024-12-02 at 12 19 39 p m

dyersituations (approved):

PR #486, opened 2024-12-06
feat: sidesheet expand effect

christian14b:
Adding an expand effect for sidesheet. Mainly thought for `mobile` in Homepage Borrower Profile.

https://github.com/user-attachments/assets/59bb861b-5cf9-4d76-9f44-989fbc36ebcf

### DIFF
--- a/@kiva/kv-components/vue/KvSideSheet.vue
+++ b/@kiva/kv-components/vue/KvSideSheet.vue
@@ -148,13 +148,18 @@ export default {
 
 				const rect = animationSourceElement.value?.getBoundingClientRect();
 
-				if (rect?.top) {
+				const top = rect?.top ?? 0;
+				const left = rect?.left ?? 0;
+				const width = rect?.width ?? 0;
+				const height = rect?.height ?? 0;
+
+				if (top || left || width || height) {
 					initialStyles.value = {
 						position: 'fixed',
-						top: `${rect.top}px`,
-						left: `${rect.left}px`,
-						width: `${rect.width}px`,
-						height: `${rect.height}px`,
+						top: `${top}px`,
+						left: `${left}px`,
+						width: `${width}px`,
+						height: `${height}px`,
 					};
 
 					modalStyles.value = {


@kiva/kv-components/vue/stories/KvSideSheet.stories.js
```diff
@@ -45,7 +52,33 @@ const Template = (args, {
 		) {
 			console.log(category, action, label, property, value);
 		},
+		openModal(event) {
+			if (this.expandEffect) {
+				const card = event.currentTarget;
+				if (card) {
+					const rect = card.getBoundingClientRect();
```
dyersituations, line 59: I think it would make sense to have the component itself do this logic, otherwise we'd need to find the element location anytime we use this type of animation. You could add a single property, like `animationSourceElement`, and use it in place of the two properties the component currently has.

@kiva/kv-components/vue/KvSideSheet.vue
```diff
@@ -98,36 +118,86 @@ export default {
...
+		}
 
 		const closeSideSheet = () => {
 			open.value = false;
 			kvTrackFunction.value(trackEventCategory.value, 'click', 'side-sheet-closed');
+
+			if (expandEffect.value) {
+				animating.value = true;
+
+				modalStyles.value = {
```
dyersituations, line 144: I'm not sure the specific CSS issue, but the header of the side sheet is currently disappearing before the animation starts, so the animation doesn't look smooth. The header of the side sheet should stay visible during the closing animation.

@kiva/kv-components/vue/KvSideSheet.vue
```diff
@@ -98,36 +118,86 @@ export default {
...
+		};
+
 		watch(visible, () => {
 			if (visible.value) {
 				setTimeout(() => {
 					open.value = true;
 				}, '300');
+
+				if (expandEffect.value) {
+					animating.value = true;
```
dyersituations, line 171: Right before the side sheet opens, the original button disappears, the animation waits a second, then it starts. We should have this be smoother, if possible. Like can the animation start without that delay so it doesn't look like the button disappears first?

@kiva/kv-components/vue/KvSideSheet.vue
```diff
@@ -118,7 +144,34 @@ export default {
 			if (visible.value) {
 				setTimeout(() => {
 					open.value = true;
-				}, '300');
+				}, 100);
+
+				const rect = animationSourceElement.value?.getBoundingClientRect();
+
+				if (rect?.top) {
```
dyersituations, line 151: Do we need to check all of the `rect` props? Or have fallback values?

dyersituations (approved):
We might need to adjust further once we start testing with a loan card (since animation in the iOS videos is from the card, rather than the button), but I think it's fine to test with this version.

PR #489, opened 2024-12-10
feat: scroll user to original position when sheet is opening in mobile

christian14b:
if using sourceElement, scroll to original user position on browser
 
https://github.com/user-attachments/assets/a2792964-a443-462e-a151-f8edb975c8ad


dyersituations (approved):

roger-in-kiva (approved):
Nice!
This probably also work for desktop...

PR #492, opened 2024-12-17
chore: use permanent storybook link based on main branch to always sh…

mcstover:
…ow latest state.

Any concerns with this? It will show the latest in main but will not need to be updated manually. I didn't see a way to have a dynamic link to the "latest" release tags in their docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 Monorepo containing design definitions and UI components for Kiva's style guide.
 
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

emuvente (approved):
makes sense to me!

eddieferrer (approved):

PR #493, opened 2024-12-19
fix: side sheet shouldn't be full height

dyersituations:
https://kiva.atlassian.net/browse/MP-1211

### DIFF
--- a/@kiva/kv-components/vue/KvSideSheet.vue
+++ b/@kiva/kv-components/vue/KvSideSheet.vue
@@ -10,7 +10,7 @@
 		@click.self="closeSideSheet"
 	>
 		<div
-			class="tw-absolute tw-right-0 tw-h-full tw-transition-all tw-duration-300 tw-bg-white
+			class="tw-absolute tw-right-0 tw-transition-all tw-duration-300 tw-bg-white
 				tw-overflow-y-auto tw-p-2"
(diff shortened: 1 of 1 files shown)


mcstover (approved):
Guessing this prevents the side sheet from extending to the full height of it's contents. Seems like it should match the viewport height then allow scrolling within...

dyersituations:
> Guessing this prevents the side sheet from extending to the full height of it's contents. Seems like it should match the viewport height then allow scrolling within...

It's a bit complicated because of the space for the header. In testing locally, removing this full height was a decent compromise. The sheet doesn't expand forever if the parent to larger, and the content is scrollable if the parent is smaller.

roger-in-kiva:
We could probably have new loans with no lenders and no teams sections. Said that side sheet could be smaller than viewport for large screens... agree having viewport height as min height for side sheet.

dyersituations:
> We could probably have new loans with no lenders and no teams sections. Said that side sheet could be smaller than viewport for large screens... agree having viewport height as min height for side sheet.

The problem again is the nav bar, we could add JS to dynamically work with the header sizing, and maybe we'll need to do that eventually. This solution seems to be a decent compromise when I test it locally.

roger-in-kiva (approved):

PR #495, opened 2025-01-03
feat: add autoplay and fade plugin options for carousels

eddieferrer:
* Adds fade plugin -- initialized via `:fade-enabled="true"`
* Adds autoplay plugin initialized via: `:autoplay-options="{ playOnInit: true, delay: 3000 }"` (for example)

See stories for use, theres some trickyness to get the auto playing status and trigger autoplay outside of the carousel, mostly related to when the carousel is initialized and the fact that most carousel interactions actually pause the auto play behavior


dyersituations (approved):

mcstover (approved):
Cool!

PR #496, opened 2025-01-04
fix: update font cdn url

emuvente:
UI and CPS are pre-loading fonts while using `https://www.kiva.org` as the cdn url. Because this config is still using the a different url (the fastly freetls url) for the cdn, the fonts are re-downloaded when the tailwind css is parsed. Making the urls match fixes this (confirmed locally with CPS).

I also found that UI is generating different font files now. This doesn't cause the same re-downloading issue, because on UI pages the tailwind css is placed first, so the UI font-face rules override the tailwind ones and the urls from the tailwind css are ignored. I think a good follow-up task to this is centralizing the font files and definitions in kv-tokens, but this quick fix will address the font re-downloading issue in CPS for now.


mcstover (approved):
Good find!

PR #497, opened 2025-01-06
feat: new lendCta variant with preset options as buttons plus dropdown

christian14b:
image

image

### DIFF
--- a/@kiva/kv-components/vue/KvLendCta.vue
+++ b/@kiva/kv-components/vue/KvLendCta.vue
@@ -61,6 +61,7 @@
 				<div
 					v-if="showPresetAmounts"
 					class="tw-flex tw-gap-1"
+					:class="{'tw-flex-wrap md:tw-flex-nowrap': enableHugeAmount}"
 				>
 					<template v-if="!isLendAmountButton">
 						<kv-ui-button
@@ -80,8 +81,12 @@
 						v-if="showFilteredDropdown"
 						:id="`LoanAmountDropdown_${loanId}`"
 						v-model="selectedDropdownOption"
-						class="tw-flex-1 tw-min-w-12 tw-rounded filtered-dropdown"
-						:class="{'unselected-dropdown': !selectedDropdown, 'selected-dropdown': selectedDropdown}"
+						class="tw-min-w-12 tw-rounded filtered-dropdown"
+						:class="{
+							'unselected-dropdown': !selectedDropdown,
+							'selected-dropdown': selectedDropdown,
+							'tw-w-full': enableHugeAmount
+						}"
 						aria-label="Lend amount"
 						@update:modelValue="trackLendAmountSelection"
 						@click.native.stop="clickDropdown"
@@ -270,6 +275,10 @@ export default {
 		showPresetAmounts: {
 			type: Boolean,
 			default: false,
+		},
+		kvTrackCategory: {
+			type: String,
+			default: 'Lending',
 		},
 	},
 	data() {
@@ -475,7 +484,7 @@ export default {
 	methods: {
 		async addToBasket() {
 			this.kvTrackFunction(
-				'Lending',
+				this.kvTrackCategory,
 				'Add to basket',
 				this.showLendAgain ? 'Lend again' : 'lend-button-click',
 				this.loanId,
@@ -498,7 +507,7 @@ export default {
 			}
 
 			this.kvTrackFunction(
-				'Lending',
+				this.kvTrackCategory,
 				'Modify lend amount',
 				selectedDollarAmount,
 				this.loanId,
@@ -506,7 +515,9 @@ export default {
 			);
 		},
 		clickDropdown() {
-			this.kvTrackFunction('Lending', 'click-Modify loan amount', 'open dialog', this.loanId, this.loanId);
+			this.kvTrackFunction(
+				this.kvTrackCategory, 'click-Modify loan amount', 'open dialog', this.loanId, this.loanId,
+			);
 		},
 		clickSecondaryButton(event) {
 			if (this.secondaryButtonHandler) {
@@ -520,12 +531,14 @@ export default {
 			}
 		},
 		clickPresetButton(selectedDollarAmount) {
-			this.kvTrackFunction('Lending', 'Modify lend amount', selectedDollarAmount, this.loanId, this.loanId);
+			this.kvTrackFunction(
+				this.kvTrackCategory, 'Modify lend amount', selectedDollarAmount, this.loanId, this.loanId,
+			);
 			this.selectedOption = selectedDollarAmount;
 		},
 		handleCheckout() {
 			this.kvTrackFunction(
-				'Lending',
+				this.kvTrackCategory,
 				'click-Continue-to-checkout',
 				'Continue to checkout',
 				this.loanId,


@kiva/kv-components/vue/KvLendCta.vue
```diff
@@ -54,10 +54,54 @@
...
-				<div class="amountDropdownWrapper">
+				<div
+					v-if="showPresetAmounts"
+					class="tw-flex tw-gap-1"
+				>
+					<template v-if="!showUnreservedAmountOnly">
+						<kv-ui-button
+							v-for="option in defaultAmountOptions"
+							:key="option"
+							:state="`${ unreservedAmount < option ? 'disabled' : ''}`"
```
dyersituations, line 69: Could we add one more story for this logic, when the amount left is less than $75, but a increment of $25. So unreserved of $25 would be good, so we can see the rest of the buttons disabled.

@kiva/kv-components/vue/KvLendCta.vue
```diff
@@ -54,10 +54,54 @@
 		>
 			<fieldset
 				class="tw-w-full tw-flex"
+				:class="{'tw-flex-col tw-gap-1': showPresetAmounts}"
 				:disabled="isAdding"
 				data-testid="bp-lend-cta-select-and-button"
 			>
-				<div class="amountDropdownWrapper">
+				<div
```
dyersituations, line 61: Could we add a story for when the CTA has presets but the loan is funded?

@kiva/kv-components/vue/KvLendCta.vue
```diff
@@ -79,11 +123,12 @@
 				</div>
 
 				<!-- Lend button -->
-				<div :class="{ 'lendButtonWrapper': hideShowLendDropdown }">
+				<div :class="{ 'lendButtonWrapper': hideShowLendDropdown && !showPresetAmounts}">
```
dyersituations, line 126: Also to make sure these use cases are still covered, could you add preset stories for: loading, adding, and basketed? Note that the ticket calls out that it should say, for example, "Continue to basket".

@kiva/kv-components/vue/KvLendCta.vue
```diff
@@ -54,10 +54,54 @@
...
+						<kv-ui-button
+							v-for="option in defaultAmountOptions"
+							:key="option"
+							:state="`${ unreservedAmount < option ? 'disabled' : ''}`"
+							variant="secondary"
+							class="tw-inline-flex tw-flex-1 preset-option"
+							:class="{'selected-option': selectedButtonOption === option }"
+							data-testid="bp-lend-cta-lend-button"
+							type="submit"
+							@click="selectedButtonOption = option"
```
dyersituations, line 75: It looks like these buttons are missing new analytics. Please make sure analytics similar to the dropdown are logged on click.

@kiva/kv-components/vue/KvLendCta.vue
```diff
@@ -54,10 +54,54 @@
...
+						>
+							${{ option }}
+						</kv-ui-button>
+						<kv-ui-select
```
dyersituations, line 79: Dropdown should probably have the pointer cursor to match the buttons, since the UI of the elements look the same.

dyersituations:
> Yeah.. default buttons are 25, 50 75, but we can add some logic to get the initial 3 from prices property. thoughts @dyersituations ?

I think the comment is that we need to make sure a single button works for the situation that @roger-in-kiva mentioned. We don't currently have $5 notes, so if we have less than $25, there should just be a single button that remains.

roger-in-kiva:
> > Yeah.. default buttons are 25, 50 75, but we can add some logic to get the initial 3 from prices property. thoughts @dyersituations ?
> 
> I think the comment is that we need to make sure a single button works for the situation that @roger-in-kiva mentioned. We don't currently have $5 notes, so if we have less than $25, there should just be a single button that remains.
> 
> Basically we just need stories that show those use cases ($10, $15, $30 remaining). I think your current logic covers those cases.

Currently having a $35 case on production.
Screenshot 2025-01-07 at 11 54 26 a m

roger-in-kiva:
For these cases we should show $25 as well.
Screenshot 2025-01-07 at 2 09 52 p m
Screenshot 2025-01-07 at 2 12 27 p m

For this we should show $50 as well
Screenshot 2025-01-07 at 2 11 37 p m

Not sure if for greater than $75 we should show it in the CTA
Screenshot 2025-01-07 at 2 14 45 p m

dyersituations:
> Yeah.. I think the same.. showing 25 button, but how we get back to the unreserved amount after we click 25? For instance, the first case ? 40 should appear on the dropdown?

I think we can just enable the buttons that would work and keep the amount in the CTA rather than the dropdown.

roger-in-kiva:
> For these cases we should show $25 as well. Screenshot 2025-01-07 at 2 09 52 p m Screenshot 2025-01-07 at 2 12 27 p m
> 
> For this we should show $50 as well Screenshot 2025-01-07 at 2 11 37 p m
> 
> Not sure if for greater than $75 we should show it in the CTA Screenshot 2025-01-07 at 2 14 45 p m



> > For these cases we should show $25 as well. Screenshot 2025-01-07 at 2 09 52 p m Screenshot 2025-01-07 at 2 12 27 p m
> > For this we should show $50 and $25 as well Screenshot 2025-01-07 at 2 11 37 p m
> > Not sure if for a non $25 increment greater than $75 we should show it in the CTA Screenshot 2025-01-07 at 2 14 45 p m
> 
> Yeah.. I think the same.. showing 25 button, but how we get back to the unreserved amount after we click 25? For instance, the first case ? 40 should appear on the dropdown?

My guess is having it in next button, replace $50 with $40, or Casey's suggestion.

roger-in-kiva:
For this case, if I select $25 there is no way to back to $40, for that my suggestion to use next button...

Screenshot 2025-01-07 at 3 51 57 p m
Screenshot 2025-01-07 at 3 52 04 p m

roger-in-kiva:
Current `complete loan` feature is for cases less than $50, not sure if product really want this...

Screenshot 2025-01-07 at 3 54 08 p m

dyersituations:
> For this case, if I select $25 there is no way to back to $40, for that my suggestion to use next button...
> 
> Screenshot 2025-01-07 at 3 51 57 p m Screenshot 2025-01-07 at 3 52 04 p m

Yeah I think adding a $40 button (maybe replacing $50 and $75) is the cleanest solution.

roger-in-kiva:
> > For this case, if I select $25 there is no way to back to $40, for that my suggestion to use next button...
> > Screenshot 2025-01-07 at 3 51 57 p m Screenshot 2025-01-07 at 3 52 04 p m
> 
> Yeah I think adding a $40 button (maybe replacing $50 and $75) is the cleanest solution.

Probably has no sense to show disabled options... For $40 unreserved amount worth a try hiding $75 and dropdown, and other cases as well.

roger-in-kiva:
Regular case $25 should be preselected, for cases less equal than $50, `complete loan`, unreserved amount should be preselected; said that, in buttons or dropdown that amount should be showed as selected.

roger-in-kiva:
Nice progress!
For this case, selected should be $25:
Screenshot 2025-01-07 at 6 50 46 p m

roger-in-kiva:
I'm still seeing $80 selected on this case:
Screenshot 2025-01-08 at 10 14 52 a m

roger-in-kiva:
We should probably add the huge amount option as we are trying to replicate current behavior of LendCTA.
Screenshot 2025-01-08 at 10 40 09 a m

dyersituations:
@roger-in-kiva @christian14b the huge lending amounts is still an active experiment, so yes, we should include those values if the experiment is active, like the other version of the CTA.

roger-in-kiva:
> @dyersituations @roger-in-kiva thanks! yeah, I don't think that would be a problem. I was referring to the dropdown after user select a value like $1,000 as the image I shared. Currently we don't have that problem because there are no buttons that will reduce the space of the dropdown, but with the preset buttons we might see that problem.
> 
> That's why I suggested something like this in mobile, but I think it's too big for mobile phones to be honest. We can leave it as it is now, and then assess if we need to make some changes later
> 
> image

That suggestion looks good to me!

dyersituations:
> > @dyersituations @roger-in-kiva thanks! yeah, I don't think that would be a problem. I was referring to the dropdown after user select a value like $1,000 as the image I shared. Currently we don't have that problem because there are no buttons that will reduce the space of the dropdown, but with the preset buttons we might see that problem.
> > That's why I suggested something like this in mobile, but I think it's too big for mobile phones to be honest. We can leave it as it is now, and then assess if we need to make some changes later
> > image
> 
> That suggestion looks good to me! Just ensure we'll have enough space at the bottom of the modal as this is suppose to be sticky.

Ahhh, yeah I see what you mean @christian14b. Agreed with @roger-in-kiva that your suggestion looks good (newline)!

@kiva/kv-components/vue/KvLendCta.vue
```diff
@@ -418,6 +519,10 @@ export default {
 				this.handleCheckout();
 			}
 		},
+		clickPresetButton(selectedDollarAmount) {
+			this.kvTrackFunction('Lending', 'Modify lend amount', selectedDollarAmount, this.loanId, this.loanId);
```
roger-in-kiva, line 523: Probably we should emit this in order to track an specific event for BP.

dyersituations: Emit or pass in tracking category with default of `Lending`

roger-in-kiva: A vague memory Greg asked to have `borrower-profile`...

dyersituations: Right, we'd pass in `borrower-profile` for the new usage, but the existing CTA component usage still has `Lending`, hence having that be the default of the new category prop (if we go that route instead of emitting).

roger-in-kiva (approved):
🤙🏻

dyersituations (approved):

PR #498, opened 2025-01-08
feat: add autoplay button if autoplay plugin is enabled

eddieferrer:
Adds play/pause button to vertical carousel.
Modify carousel functionality a bit to allow for:

- the play/pause button to work
- the ability to play/pause from outside the carousel
- the ability to read the play/pause state from outside the carousel

Mobile alignment of buttons:
Screenshot 2025-01-08 at 12 08 51 PM
Screenshot 2025-01-08 at 12 08 46 PM
Desktop alignment of buttons:
Screenshot 2025-01-08 at 12 08 34 PM

I did not add the play/pause button to the regular carousel. I'm not sure we have a design of where it should go.

### DIFF
--- a/.github/workflows/run-ui-tests.yml
+++ b/.github/workflows/run-ui-tests.yml
@@ -17,9 +17,9 @@ jobs:
     - name: Use Node.js
       uses: actions/setup-node@v2
       with:
-        node-version: '20.x'
+        node-version: '18.x'
     - name: Install dependencies
-      run: npm ci --prefer-offline
+      run: npm ci
     - name: Build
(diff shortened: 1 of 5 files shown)


dyersituations (approved):
Approved pending action error. We did start encountering a different error recently where I had to ensure an action ran at a newer version of node, but the error in this PR is different...

mcstover (approved):
I don't think we need another storybook build for this to merge...

mcstover:
Looks like that revert worked! Didn't realize it was actually causing an issue.

PR #499, opened 2025-01-08
feat: add variation to sidesheet for sticky controls

christian14b:
In order to set LendCta cinoibebt on BP modal, `KvSidesheet` needed to have a controls section just as `kvLightbox`

image

### DIFF
--- a/@kiva/kv-components/vue/KvSideSheet.vue
+++ b/@kiva/kv-components/vue/KvSideSheet.vue
@@ -15,7 +15,6 @@
 			:class="{
 				'tw-w-0 tw-delay-200 tw-opacity-0': !open,
 				'lg:tw-w-1/2 tw-w-full tw-opacity-full': open,
-				'tw-h-screen': stickyControls,
 			}"
 			:style="modalStyles"
 		>
@@ -129,6 +128,7 @@ export default {
 			kvTrackFunction,
 			trackEventCategory,
 			animationSourceElement,
+			stickyControls,
 		} = toRefs(props);
 
 		const open = ref(false);
@@ -200,8 +200,16 @@ export default {
 							transition: 'all 0.5s ease-in-out',
 						};
 					}, 10);
-				} else {
-					modalStyles.value = {};
+				} else if (stickyControls.value) {
+					let sheetHeight = 100;
+					const navElement = document.querySelector('nav'); // Check if header element exists
+					if (window) {
+						sheetHeight = window.innerHeight - (navElement?.offsetHeight ?? 0);
+					}
+
+					modalStyles.value = {
+						height: `${sheetHeight}px`,
+					};
 				}
 			}
 		});


@kiva/kv-components/vue/KvSideSheet.vue
```diff
@@ -15,6 +15,7 @@
 			:class="{
 				'tw-w-0 tw-delay-200 tw-opacity-0': !open,
 				'lg:tw-w-1/2 tw-w-full tw-opacity-full': open,
+				'tw-h-screen': stickyControls,
```
dyersituations, line 18: Will this work with the website header?

roger-in-kiva: Should have min height?

dyersituations: I think we have two options, both somewhat complex and needing to watch for scrolling to change the UI (like we do for the home page tiles on scroll):
1. Add a prop to this component for navigation height, have the side sheet be full view height minus the navigation height, watch on scroll to change the height as you scroll, so the "full height" side sheet allows room for the navigation -> this is probably the cleanest
2. Have the controls stick to the bottom of the screen until you scroll down past the controls (basically when you scroll past the height of the side sheet)

roger-in-kiva: Different heights for mobile and desktop btw.

@kiva/kv-components/vue/KvSideSheet.vue
```diff
@@ -184,8 +200,16 @@ export default {
 							transition: 'all 0.5s ease-in-out',
 						};
 					}, 10);
-				} else {
-					modalStyles.value = {};
+				} else if (stickyControls.value) {
```
dyersituations, line 203: When does this `else` get hit? I'm confused why we need this in the `else`.

@kiva/kv-components/vue/KvSideSheet.vue
```diff
@@ -184,8 +200,16 @@ export default {
...
+					let sheetHeight = 100;
```
dyersituations, line 204: This `100` doesn't make sense, since it could fallback to `100px`

@kiva/kv-components/vue/KvSideSheet.vue
```diff
@@ -184,8 +200,16 @@ export default {
...
+					const navElement = document.querySelector('nav'); // Check if header element exists
+					if (window) {
+						sheetHeight = window.innerHeight - (navElement?.offsetHeight ?? 0);
```
dyersituations, line 207: What about when the page scrolls? The side sheet will no longer want the `offsetHeight` but whatever part of the nav is visible subtracted.

@kiva/kv-components/vue/KvSideSheet.vue
dyersituations: So this gets called again when the side sheet is opened? We'll want to make sure the animation respects the navigation height/location.

dyersituations:
> @dyersituations @roger-in-kiva I'm moving the logic on [kiva/cms-page-server#1727](https://github.com/kiva/cms-page-server/pull/1727) for now, because I realize it can be complicated to get loan information from where the bp modal is being called `[...path.vue]`. I think handling it from `BPContent` is easier for now. I can close this PR if we are ok with it
The new solution looks good, go ahead and close this PR.

PR #502, opened 2025-01-09
refactor: update deep selectors

emuvente:
Addressing build warnings like
```
[@vue/compiler-sfc] the >>> and /deep/ combinators have been deprecated. Use :deep() instead.
```

### DIFF
--- a/@kiva/kv-components/vue/KvCommentsAdd.vue
+++ b/@kiva/kv-components/vue/KvCommentsAdd.vue
@@ -116,20 +116,20 @@ export default {
 </script>
 
 <style lang="postcss" scoped>
->>> input {
+:deep(input) {
 	@apply tw-border-t-0 tw-border-r-0 tw-border-l-0 tw-rounded-none tw-p-0 tw-h-4;
 }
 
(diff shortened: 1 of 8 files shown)


eddieferrer (approved):
Thanks for doing this!

@kiva/kv-components/vue/KvCommentsAdd.vue
```diff
@@ -1,5 +1,5 @@
 <template>
-	<div class="tw-flex tw-flex-col">
+	<div class="kv-comments-add tw-flex tw-flex-col">
```
dyersituations, line 2: Is this class needed?

dyersituations (approved):

@kiva/kv-components/vue/KvCommentsAdd.vue
emuvente: No. It's slightly more performant to use class selectors in combination with attribute selectors (which is what `:deep()` turns into) so I was thinking of trying that here, but ended up not bothering and forgot to remove the class.

PR #507, opened 2025-01-10
refactor: vite for kv-components

emuvente:
Started doing this so that KvFlag and the flag svgs could be transpiled in the library, avoiding the need for UI, CPS, or other apps to handle it. 

Most notable changes are

- upgrading from node 18 to node 22
- updating entry points for importing components. One can now either `import { KvMap } from '@kiva/kv-components';` or `import KvMap from '@kiva/kv-components/vue/KvMap';` or `import KvMap from '@kiva/kv-components/vue/KvMap.js';`
- moving the components, stories, and utilities to the `src` directory
- replacing dependencies with bundledDependencies and peerDependencies
- kv-tokens is now 100% esm for source files
- KvFlag now requires the country name to be passed in
- KvCountdownTimer now takes a deadline prop for the Date to count down to instead of a msLeft prop for the number of milliseconds remaining


mcstover:
What creates the need to have a folder and additional index for every component? Is there a way around that?

I can see the usefulness of synchronizing within this repo with or without the `.vue` extension but this still leaves a `.vue` usage for each component import in the index files. Could we just skip having to change that everywhere and allow `.vue` extensions in vite and eslint.

We could also look at using an alias in vite to resolve without a relative path...something like `#components/MyComponent` instead of `../../..`. It would of course also require a bunch file updates to implement... 🤷

emuvente:
> What creates the need to have a folder and additional index for every component? Is there a way around that?

This structure is from a guide for component libraries I was following. Each index is an entry point for the library. It allows individual fully-built components to be used without depending on the rest of the library and without needing an additional build step. We can also use those individual index files to inject type definitions, if/when we add typescript to this. If we use just the top-level index file, we might not get tree-shaking with the output and it might be harder to add type definitions. I'm still testing the consumer side of this setup, and I'll change it if I can.

> We could also look at using an alias in vite to resolve without a relative path...something like #components/MyComponent instead of ../../... It would of course also require a bunch file updates to implement... 🤷

Yeah, that's a good idea. I probably should have started with that. Find & replace will make easy work of it at least

@kiva/kv-components/src/utils/index.js
```diff
@@ -1,4 +1,5 @@
 export * from './Alea';
+export { default as Alea } from './Alea';
```
dyersituations, line 2: Is the `Alea` export also needed on the line above?

@kiva/kv-components/src/vue/KvFlag.vue
```diff
@@ -40,13 +43,21 @@ export default {
 			type: String,
 			required: true,
 		},
+		/**
+		 * Name of the country of the flag
+		* */
+		name: {
```
dyersituations, line 49: Will add this new prop break existing usages that relied on the `getNameByCode`?

dyersituations: In other words, will it be a heavy lift to add this new `required` prop the next time this package is updated in the Vue apps?

@kiva/kv-components/src/utils/index.js
emuvente: When re-exporting everything from another module, it doesn't appear to be possible to do that in one line. `export * from './Alea';` will only export the named exports, not the default, and `export { default as Alea, * } from './Alea';` is not valid es syntax. Eslint reports a parsing error: `Line 1, column 27: Unexpected token *`.

@kiva/kv-components/src/vue/KvFlag.vue
emuvente: I don't expect it to be a heavy lift since in most cases the iso code data came from a graphql Country object, so the country name could be easily added to the query.

One case where it might be a heavy lift is the phone number input in the mfa settings in UI, where there is a long list of country codes that didn't come from a graphql call. We could handle that case by giving the phone number input component a country name list, and then it's only loaded when entering a phone number instead of every time a flag is used.

dyersituations: Do you plan on tackling those changes after merging this? Luckily we aren't using `KvFlag` in CPS, where we are currently using the new side sheet and still making changes, but I could see us being requested to use the side sheet BP in MyKiva in the next few months.

emuvente: Yeah, Im planning on updating UI and CPS for all the breaking changes I introduced with this PR, including adding the country name to KvFlag usage.

dyersituations (approved):

eddieferrer (approved):

PR #509, opened 2025-01-17
feat: avoid page scroll when sidesheet open and fix sheet shown in le…

christian14b:
…ft sometimes


@kiva/kv-components/vue/KvSideSheet.vue
```diff
@@ -190,6 +199,7 @@ export default {
 			closeSideSheet,
 			goToLink,
 			modalStyles,
+			avoidBodyScroll,
```
dyersituations, line 202: Do we need to export this method now that it's just being used in the component?

dyersituations (approved):

PR #510, opened 2025-01-22
fix(kv-components): include import line for component css files

emuvente:
Vite library mode does not include imports for generated css files (see https://github.com/vitejs/vite/issues/1579), and instead requires library users to manually import every needed css file. This PR avoids that by using a custom vite plugin to replace the placeholder comment for the css file with an import line. The plugin is based on the various plugins that have already been created to deal with this issue, but tailored for our needs.

### DIFF
--- a/@kiva/kv-components/build/libCss.js
+++ b/@kiva/kv-components/build/libCss.js
@@ -1,0 +1,59 @@
+import { writeFileSync } from 'node:fs';
+import {
+	basename,
+	dirname,
+	relative,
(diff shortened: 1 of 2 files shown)


mcstover (approved):
Good stuff! I saw some others had taken a similar approach in the issue you mentioned. Thanks for putting this together!

eddieferrer (approved):

dyersituations (approved):

PR #521, opened 2025-02-20
feat(kv-shop)!: compile kv-shop with vite

emuvente:
This bundles the braintree web drop in with kv-shop to make it work in vite dev mode in consumer projects.

Created a package for `@kiva/vite-plugin-vue-lib-css` so that the plugin can be shared between kv-components and kv-shop.

Known issues:
- Typescript complains about not recognizing `@kiva/kv-components` due to kv-components not having any type declarations. This likely could be addressed by setting up kv-components for typescript and adding a type definition generation step similar to how this PR is using vite-plugin-dts.

### DIFF
--- a/@kiva/kv-shop/package.json
+++ b/@kiva/kv-shop/package.json
@@ -5,7 +5,6 @@
     "publishConfig": {
         "access": "public"
     },
-    "main": "dist/index.cjs",
     "module": "dist/index.js",
     "types": "dist/index.d.ts",
     "files": [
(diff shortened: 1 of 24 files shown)


package.json
```diff
@@ -8,7 +8,8 @@
     "@kiva/kv-tokens",
     "@kiva/kv-loan-filters",
     "@kiva/kv-shop",
-    "@kiva/kv-activity-feed"
+    "@kiva/kv-activity-feed",
+	"@kiva/vite-plugin-vue-lib-css"
```
dyersituations, line 12: Strange that the package.json has an indentation difference.

@kiva/kv-shop/src/components/KvPaymentSelect.vue
```diff
@@ -89,7 +92,7 @@ export default defineComponent({
 			paymentTypes,
 			preselectVaultedPaymentMethod,
 		} = toRefs(props);
-		const container = ref<HTMLInputElement | null>(null);
+		const container = ref<any>(null);
```
dyersituations, line 95: Why didn't `HTMLInputElement` work here?

@kiva/kv-shop/package.json
```diff
@@ -5,54 +5,56 @@
...
-        }
+            "types": "./dist/index.d.ts",
+            "import": "./dist/index.js"
+        },
+        "./components/*": "./dist/components/*.js",
+        "./components/*.js": "./dist/components/*.js"
     },
     "scripts": {
-        "build": "tsup src/*.ts --format cjs,esm --dts --clean && cp -R src/components/ dist/components/",
+        "build": "vite build",
```
dyersituations, line 22: So nice to have vue 2 gone...

eddieferrer:
Eager to test this! 
I think we can clean up these workarounds as well: https://github.com/kiva/cms-page-server/blob/ce28e36aa6a99d597b276b047f77afb533e6fc90/nuxt.config.ts#L18

package.json
emuvente: Whoops, mixed tabs and spaces

@kiva/kv-shop/src/components/KvPaymentSelect.vue
emuvente: Typescript was complaining about this type not matching the dropin option type, with a really long error message that I wasn't able to fully read. At first I thought that the mismatch was due to the options type being `HTMLElement | string`, but changing the container type to HTMLElement didn't change the reported issue, which makes sense as HTMLInputElement probably implements HTMLElement.

It seems that the HTMLElement type we have doesn't match the one that the dropin has. I couldn't easily find where either type was coming from, so I changed this to `any` to avoid the typescript error.

I think I could use `typeof DropinOptions.container` here instead though

dyersituations (approved):

eddieferrer (approved):

PR #533, opened 2025-03-17
feat: add custom message prop for pill

christian14b:
Support custom messages in pill

### DIFF
--- a/@kiva/kv-components/src/vue/KvCartPill.vue
+++ b/@kiva/kv-components/src/vue/KvCartPill.vue
@@ -37,6 +37,10 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		customMessage: {
+			type: String,
+			default: '',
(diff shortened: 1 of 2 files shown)


roger-in-kiva (approved):
As copy options are growing, wondering if we should pass that text as a content slot.

dyersituations:
@roger-in-kiva @christian14b luckily the prop should work fine unless we need HTML embedded.

roger-in-kiva:
> @roger-in-kiva @christian14b luckily the prop should work fine unless we need HTML embedded.

Agreed for now @dyersituations!

We already have 3 props to drive that copy, due that my suggestion.

```
milestonesNumber: {
			type: Number,
			default: 1,
},
isCloseNextMilestone: {
			type: Boolean,
			default: false,
},
customMessage: {
			type: String,
			default: '',
},
```

PR #534, opened 2025-03-26
feat: loader added to kv user avatar component

roger-in-kiva:
- loader added to KvUserAvatar component as image is loaded

https://kiva.atlassian.net/browse/MP-1444


dyersituations:
@roger-in-kiva do the storybook stories still work? Can we get a story for the loading version?

roger-in-kiva:
> @roger-in-kiva do the storybook stories still work? Can we get a story for the loading version?

@dyersituations yeah, stories still work. We don't need and specific story for loading version as we are removing loader when image is loaded, is kind of a fallback.

dyersituations (approved):

PR #535, opened 2025-03-28
fix: fix for braintree drop in error in local dev

eddieferrer:
I tested this in CPS by editing the file in my node modules/@kiva/kv-shop there directly, this seems to work. I also ran the build in kv-ui-elements and that also builds fine. 

@emuvente Thoughts on this? Seems a bit silly that the fix is so minor. Am I missing anything here? Is there any way to test the build version in CPS more directly?


emuvente:
Making the same changes in the CPS node_modules is the most direct way to test it, which you've already done. You could also test it with npm link, although that can behave oddly at times, especially when CPS isn't up-to-date with other kv-ui-elements packages.

I'm surprised this works too, though mainly because it seems like there's a chance that window.braintree will be undefined when we use it. However, looking at the braintree code it seems like it actually waits for the script to be fully loaded in the page in its main function, and our `await import` is waiting on the result of that function, so I can see how this simple fix would work.

PR #536, opened 2025-04-02
feat: set isLoading to false when image is already loaded

christian14b:
https://kiva.atlassian.net/browse/MP-1457

- Fix avatar still in loading state when image is cached

### DIFF
--- a/@kiva/kv-components/src/vue/KvUserAvatar.vue
+++ b/@kiva/kv-components/src/vue/KvUserAvatar.vue
@@ -50,6 +50,7 @@
 			/>
 			<img
 				v-show="!isLoading"
+				ref="imageRef"
 				:src="lenderImageUrl"
 				alt="Image of lender"
 				class="tw-rounded-full tw-inline-block"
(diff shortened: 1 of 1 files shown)


roger-in-kiva:
Whats difference using `@load` and how could be this related with size issue?

dyersituations:
> Whats difference using `@load` and how could be this related with size issue?

I'm assuming `@load` doesn't get called if the image is pulled from cache?

dyersituations:
> We're currently overwritting images and placeholder sizes on UI component, so the change should be happening there. Still I'm unable to get that odd state but I'll keep trying with this in place

Oh right, I missed that this was the component library.

roger-in-kiva:
> > We're currently overwritting images and placeholder sizes on UI component, so the change should be happening there. Still I'm unable to get that odd state but I'll keep trying with this in place
> 
> Oh right, I missed that this was the component library.

Said that, worth a try adding more size variants on this component to stop overwriting...

PR #537, opened 2025-04-03
fix: remove important from placeholder size classes

christian14b:
- removing the !important from placeholder, this causing conflicts in thanks page overwritting size
- onMounted fix didn't work as expected, passing imageRef to handle if image is present

### DIFF
--- a/@kiva/kv-components/src/vue/KvUserAvatar.vue
+++ b/@kiva/kv-components/src/vue/KvUserAvatar.vue
@@ -141,7 +141,7 @@ export default {
 		});
 
 		const onImgLoad = () => {
-			// isLoading.value = false;
+			isLoading.value = false;
 		};
 
 		return {


roger-in-kiva:
I don't see placeholder in this case:
Screenshot 2025-04-03 at 11 05 13 a m

dyersituations:
> I don't see placeholder in this case: Screenshot 2025-04-03 at 11 05 13 a m

@christian14b it sounds like the default height and width aren't working with this changeset?

roger-in-kiva:
Have you tested adding !important here? https://github.com/kiva/ui/blob/72afc1451ae021d3c94d07cbbfce0b191601ba90/src/components/Thanks/BorrowerAvatarsContainer.vue#L98

dyersituations (approved):

PR #538, opened 2025-04-03
feat: wait for image to complete to avoid infinite loading in user av…

christian14b:
…atar

### DIFF
--- a/@kiva/kv-components/src/vue/stories/KvUserAvatar.stories.js
+++ b/@kiva/kv-components/src/vue/stories/KvUserAvatar.stories.js
@@ -9,9 +9,9 @@ const story = (args) => {
 	const template = (templateArgs, { argTypes }) => ({
 		props: Object.keys(argTypes),
 		components: { KvUserAvatar },
-		setup() { return { args: templateArgs, repeat: templateArgs?.repeat ?? 1 }; },
+		setup() { return { args: templateArgs }; },
 		template: `
-			<KvUserAvatar v-for="i in repeat" v-bind="args" />
+			<KvUserAvatar v-bind="args" />
 		`,
 	});
 	template.args = args;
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/src/vue/KvUserAvatar.vue
```diff
@@ -144,6 +144,25 @@ export default {
...
+						resolve(img);
+					} else {
+						setTimeout(check, 100);
+					}
+				};
+				check();
+			});
+		};
+
+		onMounted(async () => {
```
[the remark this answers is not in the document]
roger-in-kiva: Worth a try adding a story with multiple components to ensure that.

@kiva/kv-components/src/vue/KvUserAvatar.vue
```diff
@@ -144,6 +144,25 @@ export default {
 			isLoading.value = false;
 		};
 
+		const waitForImageToComplete = (img) => {
+			return new Promise((resolve) => {
+				const check = () => {
+					if (img?.complete) {
+						resolve(img);
+					} else {
+						setTimeout(check, 100);
```
dyersituations, line 153: We might want a max wait time, then fallback to empty circle. I'm wondering if there's a chance for an infinite promise here. Maybe a max wait time of 10s?

@kiva/kv-components/src/vue/KvUserAvatar.vue
```diff
@@ -144,23 +144,31 @@ export default {
...
 
 		onMounted(async () => {
 			const img = imageRef.value;
-			await waitForImageToComplete(img);
-			onImgLoad();
+			try {
+				await waitForImageToComplete(img);
+				onImgLoad();
+			} catch (err) {
+				console.log(err.message);
```
dyersituations, line 170: On error what does the avatar look like? White circle?

@kiva/kv-components/src/vue/stories/KvUserAvatar.stories.js
```diff
@@ -44,4 +44,10 @@ export const IsSmall = story({
 	isSmall: true,
 });
 
+export const Multiple = story({
+	lenderImageUrl: 'https://www.development.kiva.org/img/s100/26e15431f51b540f31cd9f011cc54f31.jpg',
```
roger-in-kiva, line 48: Ideally, to simulate what happens in TY page, we need a different image here.

dyersituations: Yeah three different images would be ideal.

@kiva/kv-components/src/vue/KvUserAvatar.vue
dyersituations: Ah so always loading, that's fine, since it shouldn't happen. I just wanted to make sure the circle didn't disappear or something.

@kiva/kv-components/src/vue/stories/KvUserAvatar.stories.js
roger-in-kiva: Also ensure adding a different key each repetition, like here: https://github.com/kiva/ui/blob/72afc1451ae021d3c94d07cbbfce0b191601ba90/src/components/Thanks/BorrowerAvatarsContainer.vue#L5

roger-in-kiva (approved):
Looks good!

PR #542, opened 2025-04-15
feat: rounded chevron option added to carousel

roger-in-kiva:
- rounded chevron option added to carousel
https://kiva.atlassian.net/browse/MP-1495

Screenshot 2025-04-15 at 3 45 21 p m

### DIFF
--- a/@kiva/kv-components/src/vue/KvCarousel.vue
+++ b/@kiva/kv-components/src/vue/KvCarousel.vue
@@ -48,7 +48,24 @@
 				:disabled="embla && !embla.canScrollPrev()"
 				@click="handleUserInteraction(previousIndex, 'click-left-arrow')"
 			>
+				<span
+					v-if="roundedChevron"
+					class="tw-inline-flex tw-w-4"
+				>
+					<svg
+						class="tw-w-full tw-h-full tw-fill-current"
+						xmlns="http://www.w3.org/2000/svg"
+						height="24px"
+						viewBox="0 -960 960 960"
+						width="24px"
+						fill="#e3e3e3"
+					>
+						<!-- eslint-disable-next-line max-len -->
+						<path d="m432-480 156 156q11 11 11 28t-11 28q-11 11-28 11t-28-11L348-452q-6-6-8.5-13t-2.5-15q0-8 2.5-15t8.5-13l184-184q11-11 28-11t28 11q11 11 11 28t-11 28L432-480Z" />
+					</svg>
+				</span>
 				<kv-material-icon
+					v-else
 					class="tw-w-4"
 					:icon="asideControls? mdiArrowLeft : mdiChevronLeft"
 				/>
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/src/vue/KvCarousel.vue
```diff
@@ -48,7 +48,24 @@
 				:disabled="embla && !embla.canScrollPrev()"
 				@click="handleUserInteraction(previousIndex, 'click-left-arrow')"
 			>
+				<span
+					v-if="roundedChevron"
+					class="tw-inline-flex tw-w-4"
+				>
+					<svg
```
mcstover, line 55: Not sure why we're diverging from the MDI Icons but if we are doing that intentionally, these svgs should be placed in a global location so they can be used elsewhere rather than embedded in this single component.

@kiva/kv-components/src/vue/KvCarousel.vue
```diff
@@ -48,7 +48,24 @@
 				:disabled="embla && !embla.canScrollPrev()"
 				@click="handleUserInteraction(previousIndex, 'click-left-arrow')"
 			>
+				<span
+					v-if="roundedChevron"
+					class="tw-inline-flex tw-w-4"
```
mcstover, line 53: Something about these styles is not properly centering the new svg. It's visually off center within the circle.

@kiva/kv-components/src/vue/KvCarousel.vue
roger-in-kiva: Yeah, there are not MDI rounded chevrons in MDI Icons and Nathan is asking to add them.
What do you suggest to have them placed in a global location?

mcstover: We probably need to start thinkng about a `kv-icons package in this repo. @dyersituations Have you all discussed any broader changes to icons on the site? We should probably get eng + design together to talk through the plans for this...

dyersituations: @mcstover this was a one-off request, we haven't seen any other new icons like this.

roger-in-kiva: There are others inline svgs across this repo, just following that path...

dyersituations: @mcstover because we don't have a set of icons, I think we might have two approaches here:
1. Allow an inline SVG for this use case
2. Push back on design to use the MDI icon
What are you thoughts on the best path forward?

mcstover: If we have a promise that this won't be used in any other interface elements, we can leave it inline. If however, other designers might start using these rounded chevrons in other places like buttons or accordians, we should extract the svg icons.

dyersituations: I'll ask, thanks.

eddieferrer:
I think we should just do these icons like they were done for the vertical carousel:

https://github.com/kiva/kv-ui-elements/blob/d31b947928eca67e11d7048f84c6a5516677cf99/%40kiva/kv-components/src/vue/KvVerticalCarousel.vue#L39

There was no svg used there. If we do them that way, and just rotate 90 degrees, they will look exactly the same and be uniform in style.

roger-in-kiva:
Closing this as finally not required, thank y'all!

PR #543, opened 2025-04-17
fix: resolve when avatar image disappears

dyersituations:
https://kiva.atlassian.net/browse/MP-1511


mcstover (approved):
odd that the image would disappear after initially loading... Suppose you could also call `onImgLoad` in the onMounted catch block...

dyersituations:
> odd that the image would disappear after initially loading... Suppose you could also call `onImgLoad` in the onMounted catch block...

It was for the avatar that moves to the ATB modal, then disappears. There is probably a more solid fix, but this was a quick solution.

PR #545, opened 2025-04-23
feat: add secondary button text with show presets

eddieferrer:
For the in-context BP in the impact dashboard, we want to use the `showPresetAmounts` and also show the secondary button text, this is not possible at the moment. 

I looked around the codebase for uses of showPresetAmounts and I believe this is a safe change to make.

### DIFF
--- a/@kiva/kv-components/src/vue/KvLendCta.vue
+++ b/@kiva/kv-components/src/vue/KvLendCta.vue
@@ -431,7 +431,7 @@ export default {
 			}
 		},
 		loanInBasketButtonText() {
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

mcstover (approved):
Was this just misplaced? `showPresetAmounts` is a prop but I didn't see any sign that it's present should change this text...`secondaryButtonText` seems appropriate.

eddieferrer:
> Was this just misplaced? `showPresetAmounts` is a prop but I didn't see any sign that it's present should change this text...`secondaryButtonText` seems appropriate.

Doesnt make much sense to me either...

PR #547, opened 2025-04-24
feat: change secondary text logic on lendcta

eddieferrer:
Ok trying this again. 

if secondaryButtonText prop value exists, always return that, else, show `'Continue to basket'` if showPresets, or `'Checkout now'` if no showPresets


mcstover (approved):
This will allow us to override the button text even when preset amounts are being used.

dyersituations (approved):

PR #548, opened 2025-04-24
feat: ignore kvuseravatar in chromatic

eddieferrer:
I noticed any PR is generating a lot of excessive diffs in chromatic. This is mostly due to KvUserAvatar using random colors for the user letter and circle background, so all stories that use that component will generate diffs in every PR. 

This will ignore the random color/backgrounds for the purposes of chromatics diff. 

Note, a diff will still be generated if the avatar changes in size

https://www.chromatic.com/docs/ignoring-elements/

### DIFF
--- a/@kiva/kv-components/src/vue/.storybook/preview.js
+++ b/@kiva/kv-components/src/vue/.storybook/preview.js
@@ -24,6 +24,7 @@ export const parameters = {
 			order: ['Base Styling', '*'],
 		},
 	},
+	chromatic: { ignoreSelectors: ['#kv-user-avatar'] },
 }
 
 // Listen for events from the dark mode plugin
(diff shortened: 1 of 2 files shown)


mcstover (commented):
Prob ok, though there have been many churning updates around the image loading and such... I guess I'd like to know that functionality is stable before excluding this one...

dyersituations:
> Prob ok, though there have been many churning updates around the image loading and such... I guess I'd like to know that functionality is stable before excluding this one...

Avatar changes seem okay so far... The changes are in prod and we've fixed the issues we identified on dev.

dyersituations:
If we want stable colors, another option could be to provide some optional props.

mcstover (approved):

PR #549, opened 2025-04-25
feat: initial work to add additional possible filters coming through …

mcstover:
…campaign saved searches

There are several new potential filter values that could be present in the category/saved-search associted with campaign contexts. In order to continue using our loan filter interfaces, these definitions need to exist when the loanSearchState validation method is called.

Note: I think I've got all the new definitions here but am still working on some additional tests.

### DIFF
--- a/@kiva/kv-loan-filters/src/loanAmount.ts
+++ b/@kiva/kv-loan-filters/src/loanAmount.ts
@@ -1,0 +1,59 @@
+import { getMinMaxRangeFilter, createMinMaxRange } from './minMaxRangeUtils';
+
+/**
+ * The min risk rating value
+ */
+export const MIN = 25;
+
+/**
+ * The max risk rating value
+ */
+export const MAX = 1000000;
+
+export const facetsKey = 'loanAmount';
+
+export const stateKey = 'loanAmount';
+
+export const getUiConfig = (options) => ({
+	type: undefined,
+	hasAccordion: undefined,
+	topLine: false,
+	bottomLine: false,
+	title: undefined,
+	shouldDisplayTitle: false,
+	itemHeaderKey: undefined,
+	placeholder: undefined,
+	facetsKey,
+	stateKey,
+	eventAction: undefined,
+	allOptionsTitle: undefined,
+	valueMap: undefined,
+	isPercentage: false,
+	displayedUnit: undefined,
+	...options,
+});
+
+export default {
+	// eslint-disable-next-line @typescript-eslint/no-unused-vars, @typescript-eslint/no-explicit-any
+	getOptions: (allFacets: any = {}, filteredFacets: any = {}) => ({ min: MIN, max: MAX, step: 25 }),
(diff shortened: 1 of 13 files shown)


eddieferrer (approved):

dyersituations (commented):
Looks good, but a couple modules are missing tests.

mcstover:
Ok additional tests have been added to cover all new filter options.

@dyersituations As I was working in these I realized that we are using a very lienient form of `any` on all FLSS filter sets. While this matches our current state in the campaign world and is probably appropriate for our general public lending filters, I believe we'll need to update the methods throughout the LoanSearchState chain to support the broader options available in FLSS. If you've done some thinking on this I'd love to chat some about that and how we might move to a more FLSS centric mangement of filter state.

@kiva/kv-loan-filters/src/loanAmount.ts
```diff
@@ -0,0 +1,59 @@
+import { getMinMaxRangeFilter, createMinMaxRange } from './minMaxRangeUtils';
+
+/**
+ * The min risk rating value
```
meganmcmillan, line 4: copy pasta

dyersituations (approved):

meganmcmillan (approved):

@kiva/kv-loan-filters/src/loanAmount.ts
mcstover: good catch. I'm also open to suggestions on the min max loanAmount here. I figured anything we pass through this code a present would fit within these bounds.

mcstover:
> > As I was working in these I realized that we are using a very lienient form of `any` on all FLSS filter sets.
> 
> @mcstover at first I thought you were referring to TS `any`, and I've always wanted to make the typing more strict, instead of just copying and pasting all methods, but I've never had the time back. The modules were originally created in `ui` without TS. Anyways... For the FLSS filtering, that's an interesting aspect to allow configurable filtering types. We could pretty simply add this as an optional parameter to `getFlssFilter` with a default of `any`. This is an example of a great use of TS long-term for these modules.

Oh yes, sorry for that confusion. I was more specifically speaking to converting filter values into an `{ any: [x, y, z]}` structure where `any` could be `eq`, `ne` or `none` among others... One of my initial thoughts was to begin allowing a pre-formatted FLSS filter value to be passed through and validated and just "reused" in the flss filter method.

@kiva/kv-loan-filters/src/loanAmount.ts
meganmcmillan: I think this is a good range. We could still theoretically go down to 5, but 25 is our regular note size so I think that is fine. 
I think our max LSE loan has been 500,000, and there have only been 7 of these. I think I've heard that that amount is a little on the high end for lenders so I don't see us going too far beyond this. We can set the max here, or a little higher. 1,000,000 will definitely cover it.

PR #551, opened 2025-04-25
fix: add new filters to filterUtils and fix facet key on trusteeId

mcstover:
🤦  This is required for the new filters to be available to the getValidatedSearchState methods. For the time being, none fo these will be exposed in the UI so I've kept them out of UI configs.


mcstover:
Noting I've also alphabetized the imports and properties. @dyersituations Please let me know if that could have any adverse downstream affects.

dyersituations (approved):

PR #552, opened 2025-04-25
feat: colocating loan card fragments and calculate unreservedAmount

emuvente:
This uses the strategy of [colocating fragments](https://www.apollographql.com/docs/react/data/fragments#colocating-fragments) to make it easier to write queries for our loan cards. The required data can be defined along with the components without knowing the overall query used to get the data, and UI & CPS can include these fragments in their queries without needing to know all the required fields.

As part of supporting that, I moved the calculations for fundraisingPercent and unreservedAmount into loanCard util in this repo so that UI and CPS don't have to each implement local resolvers that calculate those values.


mcstover (approved):
Looks good. So, now we'll be able to import these fragments for uses in other contexts and/or queries when we need the associated data!

Having the queries nested within component files seems a little odd at first but it also moves us toward a topic specific "module" that holds utils and information for that part of the loan card.

dyersituations (approved):

PR #556, opened 2025-04-28
feat: add a prop to limit amounts shown in presets

eddieferrer:
Add yet another prop to this component. 

If max amount is set, only display amounts less than that in the presets. Examples: 

Screenshot 2025-04-28 at 3 23 34 PM

___________________________________________________________________________________
Screenshot 2025-04-28 at 3 23 44 PM

### DIFF
--- a/@kiva/kv-components/src/vue/KvLendCta.vue
+++ b/@kiva/kv-components/src/vue/KvLendCta.vue
@@ -346,6 +346,14 @@ export default {
 			type: Function,
 			default: undefined,
 		},
+		/**
+		 * Hide preset buttons under this amount when
+		 * showPresetAmounts is true
+		 * */
+		maxAmount: {
+			type: String,
+			default: '',
+		},
 		showPresetAmounts: {
 			type: Boolean,
 			default: false,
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/src/vue/KvLendCta.vue
```diff
@@ -519,9 +532,12 @@ export default {
 				|| this.isAmountBetween25And500(this.unreservedAmount));
 		},
 		isLendAmountButton() {
-			return ((this.lendButtonVisibility || this.state === 'lent-to')
+			if (!this.maxAmount) {
+				return ((this.lendButtonVisibility || this.state === 'lent-to')
 				&& this.isAmountLessThan25(this.unreservedAmount))
 				|| (this.showPresetAmounts && this.presetButtonsPrices.length === 1 && !this.isAdding);
+			}
+			return false;
```
dyersituations, line 540: Why do these two need to return `false` when a max amount is used?

eddieferrer: We want the behavior shown in my screenshots, if these 2 arent manually set to false, we get a lend button displayed: "Lend $<UnreservedAmount>"  for the cases when maxAmount is only enough for 1 button (maxAmount: '25')

dyersituations (approved):

PR #557, opened 2025-04-30
feat: use Alea to get consistent 'random' value for loan callouts and user avatars

emuvente:
Making these methods use consistent 'random' values avoids client-side hydration mis-matches caused by different random values returned on the server and the client. This will also make the stories using loan callouts and user avatars consistent, so they will no longer show up as false-positives in chromatic.

The end result is that users will have a consistent avatar instead of it changing on every page load, and loan cards will only display different random callouts when their fundraising progress changes.

### DIFF
--- a/@kiva/kv-components/src/utils/imageUtils.js
+++ b/@kiva/kv-components/src/utils/imageUtils.js
@@ -1,3 +1,5 @@
+import Alea from './Alea';
+
 /**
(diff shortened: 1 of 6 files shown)


mcstover (approved):
Neat solution!

eddieferrer (approved):
Nice. We can remove these changes: https://github.com/kiva/kv-ui-elements/pull/548/files

PR #562, opened 2025-05-07
feat: use stranded lend button when max amount is 25

eddieferrer:
We want to go back to showing "Lend $25" when max amount is 25 and loan has more than 25 left unreserved

Screenshot 2025-05-07 at 10 22 15 AM


mcstover (approved):
isLendAmountButton is getting gnarly!

mcstover (approved):

PR #563, opened 2025-05-07
feat: loan channel map file updated

roger-in-kiva:
- loan channel map file updated for almost funded category


mcstover (approved):
Noting this should also be updated in the prod and dev dbs in order to keep this in sync with the db. I think we can do that without changing the Loan Channel name but am not certain...

roger-in-kiva:
> Noting this should also be updated in the prod and dev dbs in order to keep this in sync with the db. I think we can do that without changing the Loan Channel name but am not certain...

Agreed! I've added a ticket to update prod DBs when releasing this and ui repos.

PR #565, opened 2025-05-20
fix: improve SideSheet component for BadgeJourneys

The-Best-Joke:
Adds a new prop to the SideSheet component to allow for the Headline border to be hidden if specified. Also reduces the y padding on the headline slightly to conform to latest Figma designs.

### DIFF
--- a/@kiva/kv-components/src/vue/KvSideSheet.vue
+++ b/@kiva/kv-components/src/vue/KvSideSheet.vue
@@ -101,7 +101,7 @@ import { ref, toRefs, watch } from 'vue';
 import {
 	mdiClose, mdiArrowLeft, mdiExportVariant,
 } from '@mdi/js';
-import { KvMaterialIcon } from '@kiva/kv-components';
+import KvMaterialIcon from './KvMaterialIcon.vue';
 
 export default {
 	name: 'KvSideSheet',
--- a/@kiva/kv-components/src/vue/KvSideSheet.vue
+++ b/@kiva/kv-components/src/vue/KvSideSheet.vue
@@ -104,7 +104,6 @@ import {
 import KvMaterialIcon from './KvMaterialIcon.vue';
 
 export default {
-	name: 'KvSideSheet',
 	components: {
 		KvMaterialIcon,
 	},
--- a/@kiva/kv-components/src/vue/KvSideSheet.vue
+++ b/@kiva/kv-components/src/vue/KvSideSheet.vue
@@ -191,6 +191,7 @@ export default {
 		const modalStyles = ref({});
 		const controlsRef = ref(null);
 		const headlineRef = ref(null);
+		const windowHeight = ref(window.innerHeight);
 
 		const heights = reactive({
 			headline: 0,
@@ -198,8 +199,8 @@ export default {
 		});
 
 		const contentHeight = computed(() => {
-			const height = window.innerHeight - heights.headline - heights.controls;
-			return Math.max(height, 0); // Ensure non-negative height
+			const height = windowHeight.value - heights.headline - heights.controls;
+			return Math.max(height, 0);
 		});
 
 		// Debounce function to limit rapid ResizeObserver calls
@@ -212,7 +213,7 @@ export default {
 		};
 
 		const updateHeights = () => {
-			// Use setTimeout to wait for DOM rendering and styling
+			windowHeight.value = window.innerHeight;
 			setTimeout(() => {
 				nextTick(() => {
 					if (headlineRef.value) {
@@ -221,7 +222,6 @@ export default {
 					} else {
 						heights.headline = 0;
 					}
-
 					if (slots.controls?.() && controlsRef.value) {
 						const controlsRect = controlsRef.value.getBoundingClientRect();
 						heights.controls = controlsRect.height;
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/src/vue/KvSideSheet.vue
```diff
@@ -70,7 +71,7 @@
 					</div>
 				</div>
 				<div
-					class="tw-p-4 tw-overflow-y-auto tw-transition-opacity tw-duration-500 tw-delay-200
+					class="tw-p-2 tw-overflow-y-auto tw-transition-opacity tw-duration-500 tw-delay-200
```
dyersituations, line 74: I wonder if the padding should be `tw-p-2 lg:tw-p-4`

@mcstover @eddieferrer would this work for the impact dashboard use case?

mcstover: From what I see this change will result in even more cut off content when controls are present.

image

I think we mainly need to make sure the bottom of the content is reachable/visible when controls are present. I'm not concered with the left/right padding changes. The top content starts below the header so it's ok. The bottom content is getting covered/obstructed.

dyersituations: Good to know that the padding is less of a concern. @The-Best-Joke thoughts on the bottom overlap? I know that is a bit beyond what you're currently working on, but curious if you have ideas on potential fixes.

The-Best-Joke: @dyersituations Nah, this is very much on-point for the ticket I think. My solution entirely modularized to the bit of code in the `ui` repo revolved around hacking the overflow value for the contents inside the `#default` and `#control` templates so that they played nice, but I did always feel it was a bit too hacky.

I think a proper solution would be to actively modify the `SideSheet` component itself so that it always manages to fit all content inside of its `#default` template, since that is a very reasonable expectation to have regarding the use of the component by any future developer, as opposed to having to hack your way around it like I did.

A simple padding that offsets the height of the content whenever the `#control` template is present ought to do it.

dyersituations: @The-Best-Joke do you it would be worth porting that fix to this branch? If it isn't more than a few hours, I think it would be worth it.

@kiva/kv-components/src/vue/KvSideSheet.vue
```diff
@@ -1,3 +1,4 @@
+```vue
```
dyersituations, line 1: Remove.

@kiva/kv-components/src/vue/KvSideSheet.vue
```diff
@@ -69,24 +67,31 @@
...
+				<!-- Flex container with dynamic padding-bottom -->
 				<div
-					class="tw-p-4 tw-overflow-y-auto tw-transition-opacity tw-duration-500 tw-delay-200
-						tw-overscroll-y-contain"
-					:class="{
-						'tw-opacity-0': !open,
-						'tw-opacity-full': open,
-					}"
+					class="tw-flex-1 tw-overflow-y-auto tw-overscroll-y-contain"
+					:style="{ paddingBottom: controlsHeight + 'px' }"
```
dyersituations, line 73: Thoughts on possibility of having this change the component height instead of padding? Currently the scrollbar disappears/overlaps with the control footer, scrolls behind the control component.
image

The-Best-Joke: That can also be done fairly easily I think. I'll report back.

@kiva/kv-components/src/vue/KvSideSheet.vue
```diff
@@ -188,17 +186,15 @@ export default {
...
+			let controlsHeight = 0;
 			if (slots.controls?.() && controlsRef.value) {
-				const rect = controlsRef.value.getBoundingClientRect();
-				controlsHeight.value = rect.height; // Includes padding and margins
-			} else {
-				controlsHeight.value = 0; // Reset when controls slot is absent
+				const controlsRect = controlsRef.value.getBoundingClientRect();
+				controlsHeight = controlsRect.height;
 			}
+			contentHeight.value = window.innerHeight - controlsHeight;
```
dyersituations, line 197: Here we also need to subtract the height of the header.

The-Best-Joke: I honestly thought so too but local testing revealed that it left an awkward white space at the bottom of the content block that disappeared when I took out the header height. I'll test it on the Impact Dashboard to see if it also works there.

dyersituations: When in test locally, I see the content cut off and the scrollbar behind the controls:
image

@kiva/kv-components/src/vue/KvSideSheet.vue
```diff
@@ -236,22 +273,33 @@ export default {
...
 				onUnmounted(() => controlsObserver.disconnect());
 			}
-			window.addEventListener('resize', updateContentHeight);
-			onUnmounted(() => window.removeEventListener('resize', updateContentHeight));
+			if (headlineRef.value) {
+				const headlineObserver = new ResizeObserver(debouncedUpdateHeights);
+				headlineObserver.observe(headlineRef.value);
+				onUnmounted(() => headlineObserver.disconnect());
+			}
+			window.addEventListener('resize', debouncedUpdateHeights);
```
dyersituations, line 293: In storybook I'm seeing a good height when the side sheet opens, but I'm not seeing the height update as the window size changes:

@kiva/kv-components/src/vue/KvSideSheet.vue
```diff
@@ -144,23 +164,21 @@ export default {
 		},
 		/**
 		 * Source element position for expand animation
-		 */
+		 * */
```
dyersituations, line 167: I don't know why/when we started adding these extra asterisks to the jsdocs in this repo... Haha. No change needed, just something funny.

dyersituations (approved):

dyersituations:
@The-Best-Joke changes look good! The last step is for you to open the Chromatic UI Tests pending check, sign in with your GitHub, and approve manually. We still have some components that show changes with each PR...

PR #572, opened 2025-05-28
feat: add widthDimensions prop to KvSideSheet

The-Best-Joke:
Adds the `widthDimensions` prop to the `KvSideSheet` component in order to bring the component fully in line to Figma mock standards.

In the process it also slightly modifies the transition timings in order to improve the out transition.

### DIFF
--- a/@kiva/kv-components/src/vue/KvSideSheet.vue
+++ b/@kiva/kv-components/src/vue/KvSideSheet.vue
@@ -10,15 +10,14 @@
 		@click.self="closeSideSheet"
 	>
 		<div
+			ref="sideSheetRef"
 			class="tw-fixed tw-right-0 tw-transition-all tw-duration-300 tw-bg-white tw-overflow-y-auto"
-			:class="[
-				{
-					'tw-w-0 tw-delay-200 tw-opacity-0': !open,
-					'tw-opacity-full': open,
-					'tw-h-full': $slots.controls,
-				},
-				widthDimensions && open ? widthDimensions : 'lg:tw-w-1/2 tw-w-full'
-			]"
+			:class="{
+				'tw-w-0 tw-delay-200 tw-opacity-0': !open,
+				'tw-opacity-full': open,
+				'tw-h-full': $slots.controls,
+			}"
+			:style="sideSheetStyles"
 		>
 			<div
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/src/vue/KvSideSheet.vue
```diff
@@ -176,6 +181,14 @@ export default {
 			type: String,
 			default: '',
 		},
+		/**
+		 * The tailwind string width dimensions of the SideSheet
+		 * Can accept tailwind responsive designs
+		 * */
+		widthDimensions: {
```
dyersituations, line 188: We've generally stayed away from Tailwind JIT classes, due to them populating our compiled CSS file for Tailwind. Could another option be dynamic CSS in Vue? For example:

```
<style lang="postcss" scoped>
.side-sheet-container {
    width: v-bind('width');
}
</style>
```

No worries if that doesn't work here, but it might be worth investigating to keep down the size of our compiled Tailwind file.

@kiva/kv-components/src/vue/KvSideSheet.vue
```diff
@@ -176,6 +181,25 @@ export default {
 			type: String,
 			default: '',
 		},
+		/**
+		 * The CSS width of the SideSheet, either as a string (e.g., '33.33%') or an object with
+		 * breakpoint-specific widths (e.g., { default: '100%', lg: '50%' }). Supports responsive
+		 * design with breakpoints: default, sm, md, lg, xl, 2xl
+		 * */
+		widthDimensions: {
```
dyersituations, line 189: Please add a story that passes in some breakpoint values so we can see it live in storybook.

@kiva/kv-components/src/vue/KvSideSheet.vue
The-Best-Joke: @dyersituations It works!

…

But it's required the addition of a fair bit of code in order to replace Tailwind's responsive designs. Let me know which version is preferable.

dyersituations: I think the new solution looks good and will allow us to keep our compiled CSS size down. Thanks for investigating!

mcstover (approved):

PR #575, opened 2025-06-02
fix: fix the KvSideSheet closing animation

The-Best-Joke:
Fixes the closing animation for the `KvSideSheet` component by creating a dynamically determined animation width variable, necessary as the determined width of the component depends on both the current width of the viewport as well as the `widthDimensions` component prop.

Also made sure not to break things such as the resizing of the component once already opened.

### DIFF
--- a/@kiva/kv-components/src/vue/KvSideSheet.vue
+++ b/@kiva/kv-components/src/vue/KvSideSheet.vue
@@ -13,8 +13,8 @@
 			ref="sideSheetRef"
 			class="tw-fixed tw-right-0 tw-transition-all tw-duration-300 tw-bg-white tw-overflow-y-auto"
 			:class="{
-				'tw-w-0 tw-delay-200 tw-opacity-0': !open,
-				'tw-opacity-full': open,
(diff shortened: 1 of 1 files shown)


dyersituations:
Two issues I'm seeing in the stories:
1. Some stories have the side sheet get larger then smaller with the animation - is this expected?
2. The story with the controls shows the controls sliding down from the top as the side sheet animations - they should remain at the bottom, is possible.

The-Best-Joke:
@dyersituations https://github.com/kiva/kv-ui-elements/pull/575#issuecomment-2932229189

That's interesting… The current fix includes a solution precisely to avoid that exact behavior, and on the actual site the issue is not present. It's interesting that it is re-appearing in Storybook.

I'll look into it. It's likely that the way Storybook is rendering the component on that first initial load is hard-wiring the incorrect width value within their library for any subsequent call to the component, and because the component is very watch dependent, the functionality is not properly making itself be felt within the constraints of Storybook.

dyersituations (approved):

PR #584, opened 2025-06-11
feat: add new secondary nav component 

tomniftic:
this branch introduces a new “KvSecondaryNav” component with sticky scrolling outlined in [this proposal ](https://kiva.atlassian.net/wiki/spaces/LW/pages/3730833415/Proposal+Contentful+Secondary+Menu)

In it's current state the component takes a few params, we will probably need to modify it a bit once the contentful fields/setup is complete but the current params are:
- Heading - The heading text, hidden for >= medium viewport size if linkAlignment is left/center
- linkAlignment - left/right/center alignment for the links section on browsers >= medium viewport size
- Links
- - Title - link title
- - href - link location
- - isExternal - if link should render as an <a> or <router-link> 
- - isActive - if link should get active styling (underline)
- Theme - default/dark which adjusts the style passed to kv-theme-provider

### DIFF
--- a/@kiva/kv-components/src/vue/KvSecondaryNav.vue
+++ b/@kiva/kv-components/src/vue/KvSecondaryNav.vue
@@ -1,0 +1,209 @@
+<template>
+	<kv-theme-provider
+		:theme="themeStyle"
+		class="kv-tailwind"
+	>
+		<div
+			class="tw-z-1 tw-w-full"
+			:class="{
+				'tw-fixed': isBelowElement,
+				'tw-absolute': !isBelowElement,
+			}"
+		>
+			<div class="tw-w-full tw-text-primary tw-bg-primary tw-overflow-x-auto kv-secondary-nav">
+				<div
+					ref="navInner"
+					class="
+						kv-secondary-nav__inner
+						tw-flex tw-gap-2 md:tw-gap-4 tw-items-center tw-flex-wrap
+						tw-px-2 md:tw-px-8 tw-py-2"
+					:class="navAlignmentClass"
+				>
+					<div
+						v-if="heading && heading.length > 0"
+						class="kv-secondary-nav__heading-container"
+					>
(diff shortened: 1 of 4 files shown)


eddieferrer:
I think it would be nice to emit an event when a link is clicked. I think there are cases when the parent component would want to know what link was clicked so it can set 'active'

@kiva/kv-components/src/vue/KvSecondaryNav.vue
```diff
@@ -0,0 +1,212 @@
+<template>
+	<kv-theme-provider
+		:theme="themeStyle"
+		class="kv-tailwind"
+	>
+		<div
+			class="tw-z-1 tw-w-full"
+			:class="isBelowElement ? 'tw-fixed' : 'tw-absolute'"
```
eddieferrer, line 8: It seems that when scrolling starts and the fixed class is applied the component moves a little to the right. It also breaks out of the containing element.

@kiva/kv-components/src/vue/KvSecondaryNav.vue
```diff
@@ -0,0 +1,212 @@
...
+						@click="toggleSubNavigation"
+					>
+						<kv-material-icon
+							:icon="subNavigationOpen ? mdiChevronUp : mdiChevronDown"
+						/>
+					</button>
+					<div
+						ref="subNavigation"
+						class="kv-secondary-nav__links-container tw-pt-1 md:tw-pt-0 md:tw-block tw-w-full md:tw-w-auto"
+						:class="!subNavigationOpen ? 'tw-hidden' : '' "
```
dyersituations, line 45: Doesn't need to change, but this can also be:

```
:class="{ 'tw-hidden': !subNavigationOpen }"
```

eddieferrer (approved):
Looks good, thanks for cleaning that up!

PR #588, opened 2025-06-12
fix: sidesheet content height in mobile

christian14b:
`contentHeight`computed property didn't take into account that on mobile `headlineRef` "dissapears" making its background transparent

### DIFF
--- a/@kiva/kv-components/src/vue/KvSideSheet.vue
+++ b/@kiva/kv-components/src/vue/KvSideSheet.vue
@@ -195,7 +195,10 @@ export default {
 				return false;
 			},
 		},
-		hideHeadlineBg: {
+		/**
+		 * Whether to hide the background of the headline section on mobile (e.g., in Borrower Profile Sidesheet)
+		 * */
+		hideHeadlineBgOnMobile: {
 			type: Boolean,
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

@kiva/kv-components/src/vue/KvSideSheet.vue
```diff
@@ -195,6 +195,13 @@ export default {
 				return false;
 			},
 		},
+		/**
+		 * Whether to hide the background of the headline section on mobile (e.g., in Borrower Profile Sidesheet)
+		 * */
+		hideHeadlineBgOnMobile: {
```
dyersituations, line 201: Why is this passed in as a flag? Are there different use cases for mobile?

dyersituations: Where is it hidden? Just trying to understand the differences between the use cases, but I thought they were relatively the same.

dyersituations: Ah makes sense, thanks for the example!

dyersituations (approved):

PR #591, opened 2025-06-20
feat: remove huge amount flags

roger-in-kiva:
- huge amount flags removed enabling as a feature

### DIFF
--- a/@kiva/kv-components/src/vue/stories/KvLendCta.stories.js
+++ b/@kiva/kv-components/src/vue/stories/KvLendCta.stories.js
@@ -84,7 +84,6 @@ export const HugeAmount = story({
 	},
 	unreservedAmount: '12850.00',
 	kvTrackFunction,
-	enableHugeAmount: true,
 	isVisitor: false,
 });
 
@@ -95,7 +94,6 @@ export const HugeAmountUnder1000 = story({
 	},
 	unreservedAmount: '850.00',
 	kvTrackFunction,
-	enableHugeAmount: true,
 	isVisitor: false,
 });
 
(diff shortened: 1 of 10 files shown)


@kiva/kv-components/src/vue/stories/KvLendCta.stories.js
```diff
@@ -320,7 +318,6 @@ export const WithPresetOptionsHugeAmount = story({
 	},
 	unreservedAmount: '12850.00',
 	kvTrackFunction,
-	enableHugeAmount: true,
```
dyersituations, line 323: We can remove these stories dedicated to the "huge amount".

@kiva/kv-components/src/vue/stories/KvWideLoanCard.stories.js
```diff
@@ -292,5 +292,4 @@ export const HugeLentAmount = story({
 	kvTrackFunction,
 	photoPath,
 	isVisitor: false,
-	enableHugeAmount: true,
```
dyersituations, line 295: This story can be removed.

@kiva/kv-components/src/vue/stories/KvLendCta.stories.js
roger-in-kiva: I did it, but reverted then as are the cases we have to visualize that feature, wdyt?

dyersituations: Isn't the huge lending amount just a default now in the other stories?

dyersituations: Ah that makes sense, thanks, we can keep the stories then.

dyersituations (approved):

PR #594, opened 2025-06-24
fix: loan card computed properties bug

roger-in-kiva:


@kiva/kv-components/src/vue/KvIntroductionLoanCard.vue
```diff
@@ -389,6 +389,10 @@ export default {
 			type: Boolean,
 			default: false,
 		},
+		customHref: {
```
dyersituations, line 392: Does this need to be referenced in the component as well or was the error just due to missing prop?

roger-in-kiva: I guess error is due to missing prop as utility is expecting that one:
https://github.com/kiva/kv-ui-elements/blob/7c7ce1e484373f0db5b0a587d9dd971958f9124a/%40kiva/kv-components/src/utils/loanCard.js#L71

dyersituations (approved):

PR #597, opened 2025-07-08
feat: showPill code removal

tiarascoleman:


@kiva/kv-components/src/vue/KvLendCta.vue
```diff
@@ -1,7 +1,7 @@
 <template>
 	<div class="tw-whitespace-nowrap">
 		<kv-cart-pill
```
dyersituations, line 3: We actually want to remove `KvCartPill` entirely, since we no longer use this component.

dyersituations, line 3: We actually want to remove `KvCartPill` from the CTA component. The pill component is used in the checkout page in `ui`, but it's currently used for the CTA.

dyersituations: We can also remove the `showPill` usage in `cms-page-server` where we pass in `false` on the impact dashboard.

tiarascoleman: Will remove in another brief simple ticket!

dyersituations (approved):

dyersituations (approved):

PR #598, opened 2025-07-09
fix: checkout now cta to full width

christian14b:
https://kiva.atlassian.net/browse/MP-1728

image

### DIFF
--- a/@kiva/kv-components/src/vue/KvLendCta.vue
+++ b/@kiva/kv-components/src/vue/KvLendCta.vue
@@ -46,8 +46,7 @@
 		<kv-ui-button
 			v-if="isInBasket"
 			variant="secondary"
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/src/vue/KvLendCta.vue
```diff
@@ -46,8 +46,7 @@
 		<kv-ui-button
 			v-if="isInBasket"
 			variant="secondary"
-			class="tw-inline-flex tw-flex-1"
-			:class="{ 'tw-w-full': showPresetAmounts }"
+			class="tw-inline-flex tw-flex-1 tw-w-full"
```
dyersituations, line 49: How does this look if there's progress remaining? The issue currently is when a loan is funded, so this solution would work for that case, but otherwise we still show the process bar, right?

PR #599, opened 2025-07-09
feat: set max-width to kv-select

christian14b:
https://kiva.atlassian.net/browse/MP-1569

Using a container class to use with `deep` for special cases where `KvSelect` are full width

image

### DIFF
--- a/@kiva/kv-components/src/vue/KvSelect.vue
+++ b/@kiva/kv-components/src/vue/KvSelect.vue
@@ -11,7 +11,7 @@
 				v-bind="inputAttrs"
 				v-model="inputValue"
 				:disabled="disabled"
-				class="tw-text-base tw-bg-primary tw-h-6 tw-pr-4 tw-pl-2 tw-border tw-border-tertiary tw-rounded-sm tw-appearance-none tw-w-full tw-ring-inset focus:tw-outline-none focus:tw-ring-2 focus:tw-ring-action focus:tw-border-transparent tw-text-ellipsis"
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/src/vue/KvSelect.vue
```diff
@@ -112,3 +112,13 @@ export default {
 	},
 };
 </script>
+
+<style lang="postcss" scoped>
+.container {
+	@apply tw-relative tw-w-full;
+
+	@screen md {
+		width: 192px
```
dyersituations, line 121: Do we want a width or max width?

dyersituations (approved):

@kiva/kv-components/src/vue/KvSelect.vue
```diff
@@ -112,3 +112,13 @@ export default {
 	},
 };
 </script>
+
+<style lang="postcss" scoped>
+.container {
+	@apply tw-relative tw-w-full;
+
+	@screen md {
+		max-width: 192px
```
mcstover, line 121: I don't think this should be applied globally... if there is a need for a max width we should probably do that on a context basis.

Also, why would we allow the select to be 300ish pixels on mobile but only 192 on tablet or desktop...

dyersituations: The design reqs on this one was confusing, but I just re-read the comment thread, and you're correct that designers actually didn't accept this req across the board. Looks like I forgot to change the ticket based on the comment thread.

@christian14b I think we can apply the ellipsis globally, then have the max-width set in the combo page.

I'm also seeing that there are other changes in the Figma related to the dropdown changes. Slight rounded corner changes and whatnot. Those we can also do in this changeset.

mcstover (approved):

PR #602, opened 2025-07-11
feat: update KvSelect height and weight

christian14b:
Style changes to KvSelect:
- Reducing height and setting font weight to medium

### DIFF
--- a/@kiva/kv-components/src/vue/KvSelect.vue
+++ b/@kiva/kv-components/src/vue/KvSelect.vue
@@ -11,7 +11,7 @@
 				v-bind="inputAttrs"
 				v-model="inputValue"
 				:disabled="disabled"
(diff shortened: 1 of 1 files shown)


dyersituations (approved):

dyersituations:
> @dyersituations @mcstover seems like this change might affect in other components. Not sure if we want to handle this using props. For instance:
> 
> image

Good call to test locally. I wonder if we can do this:
- change font weight here
- change dropdown height only for quick filters for now?

dyersituations (approved):

mcstover:
@christian14b @dyersituations I've accepted the results from this pr as baseline but wanted to note that while we fixed the select height here we could probably refine the text positioning a bit more. It's most obvious in the loan card select.

image

Most of the baseline changes were related to the text being a little higher within the select.

PR #609, opened 2025-07-29
feat: add round variant to kvcheckbox

infante-jaime:
Checkbox Variants: Improved support for both square and round variants, including correct icon rendering and background color logic.
Icon Zoom: Added a deep-scoped style to scale the check icon SVG (transform: scale(1.2);) for better visual appearance.
Blur on Disabled: Introduced the blurOnDisabled prop to control whether the checkbox appears blurred when disabled. Updated template logic and Tailwind classes to conditionally apply blur and opacity.
Storybook Examples: Added and clarified Storybook stories for both square and round variants, showing disabled checkboxes with and without blur, and ensured export names match Storybook URLs.

### DIFF
--- a/@kiva/kv-components/src/vue/KvCheckbox.vue
+++ b/@kiva/kv-components/src/vue/KvCheckbox.vue
@@ -5,7 +5,7 @@
 	>
 		<label
 			class="tw-inline-flex tw-items-center"
-			:class="{ 'tw-opacity-low': disabled }"
+			:class="{ 'tw-opacity-low': disabled && blurOnDisabled }"
 			:for="uuid"
 		>
 			<input
@@ -22,6 +22,7 @@
 			>
 			<!-- checkbox square background -->
 			<div
+				v-if="variant === 'square'"
 				class="
 					tw-w-3 tw-h-3 tw-mr-2
 					tw-flex-shrink-0
@@ -29,16 +30,18 @@
 					tw-border
 					tw-flex tw-justify-center tw-items-center tw-overflow-hidden
 					tw-transition-all tw-duration-100
-					peer-focus-visible:tw-ring-2 peer-focus-visible:tw-ring-action"
+					peer-focus-visible:tw-ring-2
+					peer-focus-visible:tw-ring-action
+				"
 				:class="{
-					'tw-bg-white' : !isChecked,
-					'tw-bg-action' : isChecked,
-					'tw-border-secondary' : !isChecked && valid,
-					'tw-border-action' : isChecked && valid,
-					'tw-border-danger' : !valid,
+					'tw-bg-white': !isChecked,
+					'tw-bg-action': isChecked,
+					'tw-border-secondary': !isChecked && valid,
+					'tw-border-action': isChecked && valid,
+					'tw-border-danger': !valid,
 				}"
 			>
-				<!-- checkbox icon  -->
+				<!-- checkbox icon for square -->
 				<svg
 					v-if="isChecked"
 					class="tw-w-1.5 tw-h-auto"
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/src/vue/KvCheckbox.vue
```diff
@@ -140,12 +187,14 @@ export default {
 			inputListeners,
 		} = useAttrs(context, emits);
 
+		const iconCheckCircle = computed(() => mdiCheckCircle);
```
mcstover, line 190: This shouldn't be needed. If you return mdiCheckCircle like we do in other components it will be available in the template...did you run into some issues with the basic setup?

@kiva/kv-components/src/vue/KvCheckbox.vue
```diff
@@ -162,10 +211,8 @@ export default {
 
 		const setChecked = () => {
 			if (Array.isArray(modelValue.value)) {
-				// if the model is array like <kv-checkbox v-model="['item1', 'item2']" value="item1">
 				isChecked.value = modelValue.value.includes(value.value);
 			} else {
-				// else it's a boolean like <kv-checkbox v-model="true">
```
mcstover, line 168: These are helpful, why were they removed?

@kiva/kv-components/tests/unit/specs/components/KvCheckbox.spec.js
```diff
@@ -12,96 +12,104 @@ describe('KvCheckbox', () => {
...
-		// Check that the value is `false` initially
 		expect(getByText('The checkbox value is false')).toBeDefined();
 		expect(checkboxInput.checked).toEqual(false);
 
-		// Click the checkbox and expect the value to be `true` now
 		await fireEvent.click(checkboxEl);
 		expect(getByText('The checkbox value is true')).toBeDefined();
 		expect(checkboxInput.checked).toEqual(true);
 
-		// Click the reset button and expect the value to be `false` again
```
mcstover, line 51: Same, why are these comments being removed?

@kiva/kv-components/tests/unit/specs/components/KvCheckbox.spec.js
```diff
@@ -12,96 +12,104 @@ describe('KvCheckbox', () => {
 		expect(checkboxEl).toBeDefined();
 	});
 
-	it('can\'t be toggled when the disabled prop is true', async () => {
+	it('renders round variant with a role of "checkbox"', () => {
```
mcstover, line 15: It seems like from here down we've removed tests for the default checkbox, replacing them all with tests against the round variant instead. Can we have test for both?

@kiva/kv-components/src/vue/KvCheckbox.vue
infante-jaime: yes for some reason it was not getting the icon unless I made it a computed field, will check further to try to workaround it

infante-jaime: actually is ok now

@kiva/kv-components/tests/unit/specs/components/KvCheckbox.spec.js
infante-jaime: all tests are there
coverage
 KvCheckbox.vue            |     100 |    96.87 |     100 |     100 | 279

@kiva/kv-components/tests/unit/specs/components/KvCheckbox.spec.js
infante-jaime: yes done

thank you

@kiva/kv-components/src/vue/KvCheckbox.vue
mcstover: The second comment is useful too as it describes the difference between passing an array or bool to v-model

mcstover (approved):
Looks good, thank you for "circling" back to make this a shared component variation!

dyersituations (commented):
Nice work!

PR #612, opened 2025-07-31
feat: add animation support to round variant of kvcheckbox

infante-jaime:
Some missing format details for round checkbox

### DIFF
--- a/@kiva/kv-components/src/vue/KvCheckbox.vue
+++ b/@kiva/kv-components/src/vue/KvCheckbox.vue
@@ -4,8 +4,11 @@
 		:style="styles"
 	>
 		<label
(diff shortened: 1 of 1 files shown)


mcstover:
@infante-jaime These latest updates seem ok. Do any of the existing stories exercise the problem and solution? Even before and after screenshots could be helpful to understand how this affects the circle version...

mcstover:
> @infante-jaime These latest updates seem ok. Do any of the existing stories exercise the problem and solution? Even before and after screenshots could be helpful to understand how this affects the circle version...

I suppose the chromatic diffs show this as essentially removing all the space between the circle version entries in the list...
image

image

mcstover (approved):

PR #616, opened 2025-08-05
fix(KvContentfulImg): using scale fit changes aspect ratio

emuvente:
The current default for the fit option is scale, but that can distort the image. Contentful descriptions of the fit options (emphasis mine):
- `pad`: Resize the image to the specified dimensions, padding the image if needed.
- `fill`: Resize the image to the specified dimensions, cropping the image if needed.
- `scale`: Resize the image to the specified dimensions, **changing the original aspect ratio if needed**.
- `crop`: Crop a part of the original image to fit into the specified dimensions.
- `thumb`: Create a thumbnail from the image.

It's better to crop than to scale, especially because most of our images are of people, so this changes the default to `fill`.

### DIFF
--- a/@kiva/kv-components/src/vue/KvContentfulImg.vue
+++ b/@kiva/kv-components/src/vue/KvContentfulImg.vue
@@ -184,7 +184,7 @@ export default {
 		* */
 		fit: {
 			type: String,
-			default: 'scale',
+			default: 'fill',
 			validator(value) {
 				// The value must match one of these strings
 				return ['pad', 'fill', 'scale', 'crop', 'thumb'].indexOf(value) !== -1;


eddieferrer (approved):

mcstover (approved):
Do you have any example contexts we could watch to see the difference here? Prob some blog images will look a bit different and likely better ;)

emuvente:
> Do you have any example contexts we could watch to see the difference here? Prob some blog images will look a bit different and likely better ;)

I'm not finding any at the moment. I don't recall what page I last saw this cause a problem. In many cases it's been avoided by not setting the height variable on KvContentfulImg, but then that causes other problems.

emuvente:
Actually, there is one example. [The "Giant Image" story for KvContentfulImg](https://main--608b4cf87f686c00213841b1.chromatic.com/?path=/story/kvcontentfulimg--giant-image) is currently distorted:
Screenshot 2025-08-05 at 11 54 22 AM

But it's not [in this branch](https://608b4cf87f686c00213841b1-rauuawckdg.chromatic.com/?path=/docs/kvcontentfulimg--kv-components):
Screenshot 2025-08-05 at 11 54 11 AM

PR #620, opened 2025-08-06
fix: min height class and hover color

christian14b:
Remaining style fixes


dyersituations:
> > @christian14b not sure if you're testing locally, but `npm link` should work now after recent changes from Eddie.
> 
> Isn't that change for CPS? I never tried to test kv-components against UI since CPS was much easier, but I'll try the npm link approach, thanks!

Oh you're right, it was just CPS...

PR #627, opened 2025-08-14
chore: add additional radii options and activate them in tailwind conifg

mcstover:
Adding additional radii config based on the latest design direction.

Mocked up what Storybook would look like when the token updates are released and updated in kv-components:
image


@kiva/kv-tokens/primitives.js
```diff
@@ -512,9 +512,12 @@ export default {
 	},
 	shadows: {},
 	radii: {
-		sm: 4,
+		xs: 4,
+		sm: 8,
+		md: 12,
 		default: 16,
-		lg: 40,
+		lg: 16,
```
mcstover, line 519: These updates were based on an initial scale provided by design which we can see duplciates our default as lg. I discussed making this distinct with design earlier today so we may set `lg` to something else, either 20 or 24...

mcstover:
Noting this will need to be done in 2 stages. First update kv-tokens, then update that package in kv-components in this repo and in any other repo that imports kv-tokens.

dyersituations (approved):

PR #635, opened 2025-08-19
feat: change header menu background and removing search related code

christian14b:
https://kiva.atlassian.net/browse/MP-1928
https://kiva.atlassian.net/browse/MP-1927


mcstover (approved):
The component updates within the header are probably ok though I'm not sure we want to remove the utility files...could they be in use elsewhere?

PR #638, opened 2025-08-20
feat: new impact vertical selection interface, tweaks to card frame t…

mcstover:
…o prevent odd heights.

Current State: Will add custom image handling from the contentful entry next.

image

image


@kiva/kv-components/src/vue/KvCardFrame.vue
```diff
@@ -2,10 +2,11 @@
 	<kv-theme-provider
 		:theme="themeStyle"
 		class="kv-card-frame-theme-provider"
+		:class="radiusClass"
```
mcstover, line 5: This ensures the theme provider div matches the content area preventing odd styles from custom classes applied in different scenarios.

eddieferrer (approved):

PR #645, opened 2025-08-21
feat: mp-1827 - Add New Filter to Combo Page

The-Best-Joke:
feat: add new filter to combo page

### DIFF
--- a/@kiva/kv-loan-filters/src/countriesNotLentTo.ts
+++ b/@kiva/kv-loan-filters/src/countriesNotLentTo.ts
@@ -1,0 +1,38 @@
+export const facetsKey = 'countriesNotLentTo';
+
+export const stateKey = 'countriesNotLentTo';
+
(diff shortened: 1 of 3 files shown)


roger-in-kiva:
@The-Best-Joke based on requirements I don't think we need this feature at this level, ie. we don't have an FLSS filter for this option.
So, going back to combo page level, in order to show a single pill you can add a new prop to `FilterChips` component like `customPills`, similarly we did for `excludeKeys` here:
https://github.com/kiva/cms-page-server/blob/35b6d8045d102bfbb4db855cd2fe60850c27ba18/components/loan/filter/FilterChips.vue#L66

So you can pass that pill based on `countries-not-lent-to=true` query param.
Ensure do this on created hook in order to override current filters we are prefetching.

Lemme know if you have other question.

dyersituations:
Closing since we didn't end up needing this work.

PR #656, opened 2025-08-27
fix: lend cta issue with mobile viewport

infante-jaime:
fixed some issues and refectored a bit isolated a bit more the presets code so that it woundt mess with existing variants

### DIFF
--- a/@kiva/kv-components/src/vue/KvLendCta.vue
+++ b/@kiva/kv-components/src/vue/KvLendCta.vue
@@ -156,7 +156,7 @@
 				<div
 					v-if="showPresetAmounts && !isAdding"
 					:class="{
-						'tw-flex tw-gap-1 lg:tw-gap-2 tw-flex-wrap md:tw-flex-nowrap': true,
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/src/vue/KvLendCta.vue
```diff
@@ -93,217 +99,181 @@
...
-							v-if="hideShowLendDropdown && !isLessThan25"
-							:id="`LoanAmountDropdown_${loanId}`"
-							v-model="selectedOption"
-							class="tw-min-w-12"
-							style="border-radius: 14px 0 0 14px;"
-							aria-label="Lend amount"
-							@update:modelValue="trackLendAmountSelection"
-							@click.native.stop="clickDropdown"
+					<template v-if="!isLendAmountButton">
+						<kv-ui-button
```
dyersituations, line 164: I'm noticing that comparing to the previous implementation, the desktop buttons have more padding between them. Please match the old padding, which can be seen on the deployed storybook linked below.

https://main--608b4cf87f686c00213841b1.chromatic.com/?path=/story/kvlendcta--with-preset-options-loading

Image

dyersituations:
@infante-jaime see the chromatic link in the PR, there's one case where the button is moved where it shouldn't. The other cases are the correct padding changes.

PR #659, opened 2025-08-28
fix: make KvTooltip controllable via programatic events, add stories …

mcstover:
…for placement and modifiers

See new KvTooltip stories...

### DIFF
--- a/@kiva/kv-components/src/vue/KvTooltip.vue
+++ b/@kiva/kv-components/src/vue/KvTooltip.vue
@@ -4,11 +4,14 @@
 		class="kv-tailwind"
 	>
 		<kv-popper
+			ref="popperRef"
 			:controller="controller"
 			:popper-modifiers="popperModifiers"
 			:popper-placement="placement"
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/src/vue/KvTooltip.vue
```diff
@@ -109,6 +122,27 @@ export default {
...
+		};
+
+		const handleKvPopperVisibility = (isShowing) => {
+			emit('tool-tip-visible', isShowing);
+		};
+
+		const triggerHover = (enter = true) => {
+			const element = getControllerElement();
+			if (element) {
+				const eventType = enter ? 'mouseover' : 'mouseout';
```
mcstover, line 141: Noting that I initially attempted to call methods through the ref on KvPopper but couldn't get anything to trigger or response. Thus I opted for simulating mouse events that were already initiated and handled by the KvPopper component.

emuvente (approved):
This is great!

PR #670, opened 2025-09-09
Fix mp 1988 narrow lend cta

infante-jaime:
This update adds responsive support to the KvLendCta component by tracking the actual container width with a ResizeObserver, introducing the isNarrowComponent method, and conditionally rendering the CTA button in a second row when the component is narrow, ensuring only one CTA is visible and maintaining a clean layout across all container sizes.


infante-jaime:
@mcstover I've added a new ticket to the backlog with the details needed for a long-term solution, [MP-204](https://kiva.atlassian.net/browse/MP-2047), so we can track the refactor work. Let us know if we can move forward with this change.

PR #679, opened 2025-09-22
feat: add kiva card to basket mutation and wrapper 

cfowles:
Based on this comment https://github.com/kiva/cms-page-server/pull/2349#discussion_r2363827349 – moving the mutation here instead of cms-page-server.

To test `npm run test -w @kiva/kv-shop`


dyersituations (approved):
I re-ran the tests, I think it's a random failure we get.

mcstover (approved):
Thank you!

PR #680, opened 2025-09-22
fix: add placeholder param to datepicker component 

tomniftic:


mcstover (approved):
👍

tomniftic:
> I had removed the placeholder use in #678 since it was undefined. Now, the placeholder attribute should come from the inherited attributes in `v-bind="$attrs"`

My apologies, the branch I was working on in cms-page-server didn't have the updated kv-ui-elements so I was attempting to fix the same issue that you had already fixed re: vue warnings 🤦‍♀️ I'll delete this branch now

PR #681, opened 2025-09-22
feat: Secondary nav fixes and improvements

tomniftic:
This PR contains a few fixes for the secondary nav component based on feedback received in the past, including the ability to set a link for the left aligned heading text as well as proper underline and hover states for router-links

### DIFF
--- a/@kiva/kv-components/src/vue/KvSecondaryNav.vue
+++ b/@kiva/kv-components/src/vue/KvSecondaryNav.vue
@@ -30,12 +30,20 @@
 							:class="{ 'tw-block md:tw-hidden': linkAlignment === 'left' || linkAlignment === 'center' }"
 						>
 							<component
-								:is="headingLink ? (headingLink.isExternal ? 'a' : 'router-link') : 'div'"
-								:to="headingLink ? (headingLink.isExternal ? undefined : headingLink.href) : undefined"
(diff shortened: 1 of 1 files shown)


@kiva/kv-components/src/vue/KvSecondaryNav.vue
```diff
@@ -29,9 +29,24 @@
 							class="kv-secondary-nav__heading-container"
 							:class="{ 'tw-block md:tw-hidden': linkAlignment === 'left' || linkAlignment === 'center' }"
 						>
-							<div class="kv-secondary-nav__heading tw-text-h3">
+							<component
```
dyersituations, line 32: Which link is this fix for? The issue I noticed was with the nav links themselves.

tomniftic: @dyersituations There was a feature request from the Kiva team on the Due Diligence page build where they wanted the option to set the "Due Diligence" leftmost text to have a link back to the parent page- this PR includes that change as I received that request before getting pulled in to some other work and now had time to make that adjustment with the fixes for your request

mcstover (approved):
Good updates, let us know when we can help get the contentful side of setting up the heading link in place...

dyersituations (approved):

PR #685, opened 2025-09-24
feat: fallback font styles adjusted to match primary styles

emuvente:
This PR introduces fallback font styles for PostGrotesk and Dovetail MVB with adjustments to match the primary font style in size as closely possible. This minimizes layout shift and repaints, improving performance. See https://developer.chrome.com/blog/font-fallbacks for more details

The fallback font can't match the size of the primary font perfectly, because the exact sizing adjustments differ based on the content of the text. Instead, this uses the average character width, weighted against their occurrence in English. I used the calculation provided by [Capsize](https://seek-oss.github.io/capsize/) for this PR. I also added the ability to switch which pangram is used for displaying text styles in the primitives story, so that the difference between different text content can be seen.

Screenshot 2025-09-24 at 11 32 37 AM
Screenshot 2025-09-24 at 11 41 25 AM


eddieferrer (approved):
Looks great!

PR #806, opened 2026-04-10
fix: handle gifs better in kvContentfulImg

eddieferrer:
See CIT-4083 for relevant link to slack. 

TLDR: we discovered a gif image embedded into a rich text field in a contentful page is not working: https://www.development.kiva.org/about/partner-with-us?preview=true

I discovered a couple of things:
1. If an image file is greater than 20MB, contentful ignores all query params. This was throwing me off a little since I was trying to compare the gif that was working with the one that wasn't 
2. Contentful just refuses to resize GIFs with the height/width query param. For example: 
https://www.development.kiva.org/ctfassets/images/4q5zPRN6hi0KJDTUXmcB2n/b9fd6b1b81f2c88e215f540c0fe2bb0d/loan-loop-1200w.gif?w=1000&h=1000 Just returns bad request. This, in conjunction with our retina resizing was causing issues. 
3. If the w/h param is the same as the original image, contentful will return the image since no resizing is needed, in that case, I also discovered that contentful does not support the `q` quality parameter for gifs either, and will return a bad request. 
4. for animated images webp or gif, contentful can't convert them to jpg without losing the animation (obviously)

This PR does a couple of things. 
- If an image if a GIF, just remove all the query params. We should only be using gifs for animations, and rarely in that case, the query params are just going to break everything anyway. 
- Be a little smarter about the fallback format if it is not explicitly passed in as a prop. pngs should fall back to png's and gifs should fall back to gifs, not jpgs

### DIFF
--- a/@kiva/kv-components/src/vue/stories/KvContentfulImg.stories.js
+++ b/@kiva/kv-components/src/vue/stories/KvContentfulImg.stories.js
@@ -224,3 +224,25 @@ GiantImage.args = {
 	height: 2000,
 	sourceSizes: [],
 };
+
+export const GiantGif = (args, { argTypes }) => ({
+	props: Object.keys(argTypes),
+	components: {
+		KvContentfulImg,
+	},
+	setup() { return { args }; },
+	template: `
+		<kv-contentful-img
+			alt="Descriptive alt text"
+			v-bind="args"
+		/>
+	`,
+});
+
+GiantGif.args = {
+	contentfulSrc: 'https://images.ctfassets.net/j0p9a6ql0rn7/4q5zPRN6hi0KJDTUXmcB2n/b9fd6b1b81f2c88e215f540c0fe2bb0d/loan-loop-1200w.gif',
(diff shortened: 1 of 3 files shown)


eddieferrer:
There are a bunch of uses of KvContentfulImg where we are explicitly passing the fallback-format, so there will be no change in that regard for those, but after this PR we should be able to remove that param altogether from those uses or at least some of the ones where we think there might be gifs

@kiva/kv-components/src/vue/stories/KvContentfulImg.stories.js
```diff
@@ -224,3 +224,25 @@ GiantImage.args = {
 	height: 2000,
 	sourceSizes: [],
 };
+
+export const GiantGif = (args, { argTypes }) => ({
```
mcstover, line 228: 😆 Can/Should we also add a tests to the spec file?

emuvente (approved):
Nice improvements!

mcstover (approved):

PR #822, opened 2026-05-20
fix: improve centering during and after animation to focus on active …

mcstover:
…countries.

Current behavior is re-centering on the US after completing the animation. New behavior centers on all active countries after the animation so they remain in view.


dyersituations (commented):
👍

PR #823, opened 2026-05-20
feat: mp-2831 fix datepicker stays loading

infante-jaime:
https://kiva.atlassian.net/browse/MP-2831

This issue affects KvDatePicker, which is used in the gift card scheduling flow. The component previously loaded @vuepic/vue-datepicker and its CSS through a dynamic import after mount. That works under normal conditions, but on very slow or unstable connections the async chunk can fail or stall, leaving the component stuck on Loading... with no recovery path.

The fix changes the date picker dependency to a static import. This makes the dependency part of the normal bundle path instead of a separate runtime chunk request, removing the failure mode where the date picker component mounts but its async dependency never finishes loading. The tradeoff is a small increase in bundle weight for consumers that include KvDatePicker, but for this flow the reliability improvement is worth it because a stuck date picker blocks the user from scheduling the gift card.

I also added a focused KvDatePicker test to verify that the wrapper renders the date picker and continues to pass value changes through via update:model-value and change. Since the component now statically imports the third-party CSS file, Jest needs a standard CSS mock. The styleMock.js mapping lets Jest ignore CSS imports during unit tests, which is a common pattern because Jest does not process CSS the same way the app bundler does.

### DIFF
--- a/@kiva/kv-components/jest.config.cjs
+++ b/@kiva/kv-components/jest.config.cjs
@@ -30,6 +30,7 @@ module.exports = {
 	moduleNameMapper: {
 		'^#components/(.*)$': '<rootDir>/src/vue/$1',
 		'^#utils/(.*)$': '<rootDir>/src/utils/$1',
+		'\\.css$': '<rootDir>/tests/unit/styleMock.js',
 	},
 
 	// The test environment that will be used for testing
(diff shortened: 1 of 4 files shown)


mcstover (approved):
Looks to be working well in the Storybook Publish

eddieferrer (approved):

dyersituations (approved):

PR #826, opened 2026-05-27
feat: cit 4344 kv www header basic component

mcstover:
This work is for an upcoming header experiment that places the "Partner" link into the main nav and implements the newer mobile hamburger style dropdown for About and the additional links. The approach here was to essentially duplicate the "new" global header component so we could incorporate those learning and structure here, then update the styles to more closely match our existing header in cps/ui. This also means we brought the fuse dependency in so we could cover the existing search behavior.

Stories:
- [Full Header Stories (change browser width to view mobile and tablet adaptations)](https://608b4cf87f686c00213841b1-waybywfdoc.chromatic.com/?path=/story/page-frame-kvwwwheaderbasic--logged-in) 

Todo:
- [x] Finish tabbed mobile version of hamburger menu that includes search

The work was done in concert with claude to create and operate against a plan and spec which can be found [in the ticket](https://kiva.atlassian.net/browse/CIT-4344).


eddieferrer (approved):
Looks great. I'll let Nathan chime in on visual QA if he has any small adjustments

PR #836, opened 2026-06-16
fix: ensure basket when not logged in, add hover and interaction even…

mcstover:
…ts to KvWwwHeaderBasic


eddieferrer (approved):
Looks good. Just wanted to mention that `TopNav` is the legacy category, it would be nice to convert all those categories, but thats outside the scope of this PR

PR #841, opened 2026-07-02
chore: add all color themes to kvSecondaryNav

alexdiazNiftic:
Ticket: [KUSPMTM-25](https://kiva.atlassian.net/browse/KUSPMTM-25)

Expands KvSecondaryNav to accept the full set of Kiva color themes so the mint palette in the designs can actually reach the component. Added all 11 available themes at once, same amount of work as adding one, and it avoids revisiting this if/when other themes come up.

The two hardcoded `theme === 'default'` ternaries are replaced by a themeConfig table where each theme declares both its token bundle and which background slot the nav uses (light palettes → `bg-secondary`, dark palettes → `bg-primary`).

Note on backwards compatibility: existing `theme="default"` and `theme="dark"` consumers render identically to main. Invalid theme strings also render identically. `themeStyle` returns undefined and `bgClass` falls back to `tw-bg-primary`, matching the old ternary's else branch. I kept it this way intentionally so a consumer that accidentally passes an unrecognized string doesn't silently switch from the dark-themed fallback to a default one.

[KUSPMTM-25]: https://kiva.atlassian.net/browse/KUSPMTM-25?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/@kiva/kv-components/src/vue/KvSecondaryNav.vue
+++ b/@kiva/kv-components/src/vue/KvSecondaryNav.vue
@@ -7,13 +7,13 @@
 			class="
 				tw-z-1 tw-w-full kv-secondary-nav-holder relative
 				tw-text-primary"
-			:class="theme === 'default' ? 'tw-bg-secondary' : 'tw-bg-primary'"
+			:class="bgClass"
 		>
 			<div
 				class="
 				tw-w-full tw-overflow-x-auto kv-secondary-nav
 				tw-absolute tw-top-0 tw-left-0 tw-right-0"
-				:class="theme === 'default' ? 'tw-bg-secondary' : 'tw-bg-primary'"
+				:class="bgClass"
 			>
 				<kv-page-container>
 					<div
@@ -111,7 +111,19 @@ import {
 } from 'vue';
 
 // Theme
-import { defaultTheme, greenDarkTheme } from '@kiva/kv-tokens';
+import {
+	defaultTheme,
+	greenLightTheme,
+	greenDarkTheme,
+	marigoldLightTheme,
+	stoneLightTheme,
+	stoneDarkTheme,
+	darkTheme,
+	mintTheme,
+	darkGreenTheme,
+	darkMintTheme,
+	darkStoneTheme,
+} from '@kiva/kv-tokens';
 
 // Components
 import { mdiChevronUp, mdiChevronDown } from '@mdi/js';
@@ -174,7 +186,20 @@ export default {
 			type: String,
 			default: 'default',
 			validator(value: string) {
-				return ['default', 'dark'].includes(value);
+				return [
+					'default',
+					'dark',
+					'greenLight',
+					'greenDark',
+					'marigoldLight',
+					'stoneLight',
+					'stoneDark',
+					'mint',
+					'darkGreen',
+					'darkMint',
+					'darkStone',
+					'classicDark',
+				].includes(value);
 			},
 		},
 	},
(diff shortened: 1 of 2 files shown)


@kiva/kv-components/src/vue/stories/KvSecondaryNav.stories.js
```diff
@@ -3,6 +3,25 @@ import KvSecondaryNav from '../KvSecondaryNav.vue';
...
+			control: 'select',
+			options: [
+				'default',
+				'dark',
+				'greenLight',
+				'greenDark',
+				'marigoldLight',
+				'stoneLight',
+				'stoneDark',
+				'mint',
```
mcstover: Please remove the following themes from this list: `dark`, `mint`, `darkGreen`, `darkMint`, `darkStone` and `classicDark`.  They are all deprecated and not for use.

`stoneDark` may also be deprecated but that's pending.

@kiva/kv-components/src/vue/KvSecondaryNav.vue
```diff
@@ -174,7 +186,20 @@ export default {
...
-				return ['default', 'dark'].includes(value);
+				return [
+					'default',
+					'dark',
+					'greenLight',
+					'greenDark',
+					'marigoldLight',
+					'stoneLight',
+					'stoneDark',
+					'mint',
```
mcstover: Same issue: Please remove the following themes from this list: dark, mint, darkGreen, darkMint, darkStone and classicDark. They are all deprecated and not for use.

mcstover: I have a task to remove and remap those options which will make this more explicit.

For now it may be helpful to use [this skill](https://github.com/kiva/kv-ui-elements/tree/main/%40kiva/kv-skills/design-system) to help audit and build according to what is in our approved Ecosystem Design 2026 system.

alexdiazNiftic: I'm working on the rework it based on that, just having some discussions with the design team on the component since it was using mint primary colors. Once we have that I'll update this

mcstover (approved):

@kiva/kv-components/src/vue/KvSecondaryNav.vue
```diff
@@ -207,14 +220,16 @@ export default {
...
+		const themeConfig: Record<string, { style: object; bgClass: string }> = {
+			default: { style: defaultTheme, bgClass: 'tw-bg-secondary' },
+			dark: { style: greenDarkTheme, bgClass: 'tw-bg-primary' },
+			greenLight: { style: greenLightTheme, bgClass: 'tw-bg-secondary' },
+			greenDark: { style: greenDarkTheme, bgClass: 'tw-bg-primary' },
+			marigoldLight: { style: marigoldLightTheme, bgClass: 'tw-bg-secondary' },
+			stoneLight: { style: stoneLightTheme, bgClass: 'tw-bg-secondary' },
+		};
+		const themeStyle = computed(() => themeConfig[theme.value]?.style);
+		const bgClass = computed(() => themeConfig[theme.value]?.bgClass ?? 'tw-bg-primary');
```
dyersituations, line 232: While `theme` does have a default, the default here is actually different than it was previously. It would be `tw-bg-secondary`, but it's unlikely this would ever fallback here.

PR #843, opened 2026-07-06
fix(eslint): use the vue3 lint preset instead of vue2

emuvente:
## Overview

`kv-components` and `kv-shop` both extended `plugin:vue/recommended` in their ESLint config. Due to `eslint-plugin-vue`'s legacy naming, that preset is actually the **Vue 2** ruleset. The Vue 3 equivalent is actually `plugin:vue/vue3-recommended`. Both have been linting against Vue 2 rules this whole time, flagging things Vue 3 explicitly supports (e.g. multi-root templates), while missing real Vue 3 checks entirely.

Fixing that revealed a bug and a batch of dead/stale Vue 2 patterns, so this PR also cleans those up. Getting there required first consolidating the ESLint/TypeScript/Jest tooling versions that had drifted across workspaces (see 7ea94e39f3fe5a1d6111548f0f5d926f25f7d2a8), since some packages were already ahead on `eslint-plugin-vue`/`@typescript-eslint` and some were behind.

## Commits

**`chore(deps)`** Centralize devDependencies that were declared identically (or in compatible ranges) across multiple workspaces into the root `package.json`, instead of redeclaring them per package. This eliminated a few duplicate nested `node_modules` copies caused by version drift, and unified `eslint-plugin-vue` / `@vue/eslint-config-typescript` / `@typescript-eslint` to the newer major versions.

**`fix(eslint)`** The actual preset fix: `plugin:vue/recommended` → `plugin:vue/vue3-recommended` in `kv-components` and `kv-shop`.

**`fix(kv-components)`** Everything the corrected preset caught:
- **Discovered bug**: `KvLendCta`, `KvMap`, and `KvPopper` used the deprecated `beforeDestroy` hook, renamed to `beforeUnmount` in Vue 3 (per the [migration guide](https://v3-migration.vuejs.org/breaking-changes/) and [lifecycle reference](https://vuejs.org/api/options-lifecycle.html)). Their cleanup logic was dead code. Renamed to `beforeUnmount`.
- Removed the deprecated `.native` v-on modifier (16 spots, 5 files), which is a no-op under Vue 3, since undeclared listeners already fall through as native by default.
- Added missing `emits` declarations for events that were being emitted but not declared.
- Fixed `KvTooltip`, which had `defineEmits: [...]` instead of `emits: [...]`.
- Hyphenated `@update:modelValue` → `@update:model-value` per `vue/v-on-event-hyphenation` and to match the kebab-case template syntax the [v-model guide](https://vuejs.org/guide/components/v-model.html) uses.
- Suppressed one `vue/valid-define-props` false positive in `KvSimpleMap` (confirmed via compiled build output that the flagged pattern works correctly).

**`test(kv-components)`** `KvCompactLoanCard` and `KvLendCta` had zero test coverage, so the `@update:model-value` change above had no test verifying the dropdown-selection tracking still fires. Added a focused spec per component.

## Testing

- `npm run lint` is clean across all workspaces
- `npm run test` passes
- `npm run build` succeeds across all 6 packages
- `npm run build-storybook` succeeds
- Confirmed the `beforeDestroy`/`beforeUnmount` rename against official Vue 3 docs, not just the lint rule
- Confirmed the hyphenation change matches the v-model guide's documented template syntax and red-green tested the new specs


dyersituations (approved):

eddieferrer (approved):

mcstover (approved):
Cool!

